PR #20, opened 2024-04-23
Make: use poetry to manage requirements

lusmoura:
**Description:** use poetry to manage setup instead of relying on local env


pyproject.toml
```diff
@@ -43,6 +43,11 @@ ruff = "^0.0.94"
 isort = "^5.12.0"
 black = "24.3.0"
 
+
+[tool.poetry.group.setup.dependencies]
+inquirer = "3.2.4"
+python-dotenv = "1.0.1"
```
tianjing-li, line 49: You probably need to add `python 3.11` to the dependencies, it works now because you have an existing install

KevinLuong96 (approved):

PR #29, opened 2024-04-23
tools: Update Google One-Click Deploy configuration

malexw:
The app name previously was fairly generic, so I've updated it to be more specific to the application we're deploying.

There was also an issue with the "default" field, which doesn't actually exist in the spec, and should be "value".


malexw:
This needs another round of testing before we merge

tianjing-li:
@malexw I'll be closing this since it's now duplicated by Eugenes changes

PR #49, opened 2024-04-25
coral-web: Experimental feature mode in the UI

misspia-cohere:
**Description**
Adds a banner to the page + overlay onto the settings tab when the langchain experiemental feature is turned on.
To test with this mode ON, you can add the following line to your `.env` file
```
USE_EXPERIMENTAL_LANGCHAIN=True
```
You can switch it off again by removing the line or changing the value to `False`

Screenshot 2024-04-26 at 14 39 14


src/interfaces/coral_web/src/hooks/experiementalFeatures.ts
```diff
@@ -0,0 +1,21 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { useCohereClient } from '@/cohere-client';
+
+export const useExperimentalFeatures = () => {
```
misspia-cohere, line 5: or `useIsExperimentalFeaturesMode` ?  🤔

src/interfaces/coral_web/src/hooks/experiementalFeatures.ts
```diff
@@ -0,0 +1,21 @@
...
+  const cohereClient = useCohereClient();
+
+  return useQuery<boolean>({
+    queryKey: ['experimentalFeatures'],
+    queryFn: async () => {
+      try {
+        const data = await cohereClient.listExperimentalFeatures();
+        return data.USE_EXPERIMENTAL_LANGCHAIN;
```
misspia-cohere, line 13: Feel like this is fine for now since we just have 1 experimental feature? If more of them get added, we can modify what we return based on design requirements

src/interfaces/coral_web/public/fonts/CohereIcon-Default-selection.json
wujessica, line 1: can you re-delete these :')

src/interfaces/coral_web/src/cohere-client/client.ts
```diff
@@ -45,6 +45,9 @@ export class CohereStreamError extends Error {
 
 export type Fetch = (input: RequestInfo, init?: RequestInit) => Promise<Response>;
 
+export type ExperimentalFeatures = {
+  USE_EXPERIMENTAL_LANGCHAIN: boolean;
+};
```
wujessica, line 50:
```suggestion
};

```

src/interfaces/coral_web/src/cohere-client/client.ts
```diff
@@ -318,8 +321,32 @@ export class CohereClient {
     return body as Deployment[];
   }
 
+  public async listExperimentalFeatures(): Promise<ExperimentalFeatures> {
+    const response = await this.fetch(this.getEndpoint('experimental_features'), {
```
wujessica, line 325: to save a redirect: 
```suggestion
    const response = await this.fetch(`${this.getEndpoint('experimental_features')}/`, {
```

src/interfaces/coral_web/src/components/Configuration/Settings.tsx
```diff
@@ -38,6 +40,7 @@ export const Settings: React.FC = () => {
 
   return (
     <div className="flex flex-col gap-y-6 px-5 pb-10">
+      {isExperiementalFeaturesOn && <ExperimentalModeOverlay />}
```
wujessica, line 43: correct me if im wrong but we could just delete this overlay and people would be able to interact with the elements below right? 🤣 maybe we just replace this with a ternary statement and the below component (without the `absolute` positioning)

src/interfaces/coral_web/src/hooks/experiementalFeatures.ts
wujessica: I think I'd prefer to return the entire object so that it's clearer whenever we're using the hook and also we don't need to update this in the future when we do have other flags. but if we feel strongly about this we should rename the hook to `useIsExperimentalLangchainOn`

src/interfaces/coral_web/src/components/Configuration/Settings.tsx
```diff
@@ -16,6 +17,7 @@ export const Settings: React.FC = () => {
   } = useParamsStore();
   const defaults = useSettingsDefaults();
   const { data: deployments = [] } = useListAllDeployments();
+  const { data: isExperiementalFeaturesOn } = useExperimentalFeatures();
```
wujessica, line 20:
```suggestion
  const { data: isExperimentalFeaturesOn } = useExperimentalFeatures();
```

src/interfaces/coral_web/src/pages/c/[id]/index.tsx
```diff
@@ -27,6 +29,8 @@ const ConversationPage: NextPage<Props> = () => {
   } = useParamsStore();
   const { setConversation } = useConversationStore();
   const { addCitation, resetCitations } = useCitationsStore();
+  const { data: isExperiementalFeaturesOn } = useExperimentalFeatures();
```
Comment (wujessica) [inline src/interfaces/coral_web/src/pages/c/[id]/index.tsx:32]:
```suggestion
  const { data: isExperimentalFeaturesOn } = useExperimentalFeatures();
```

src/interfaces/coral_web/src/pages/index.tsx
```diff
@@ -25,6 +27,8 @@ const ChatPage: NextPage<Props> = () => {
     setParams,
   } = useParamsStore();
   const { data: availableDeployments } = useListDeployments();
+  const { data: isExperiementalFeaturesOn } = useExperimentalFeatures();
```
wujessica, line 30:
```suggestion
  const { data: isExperimentalFeaturesOn } = useExperimentalFeatures();
```

src/interfaces/coral_web/src/cohere-client/client.ts
misspia-cohere: true nice call

src/interfaces/coral_web/src/components/Configuration/Settings.tsx
misspia-cohere: yeah that's correct haha-- i thought it'd be nice to be able to see what you have access to with experimental mode off but your suggestion is more hack proof

src/interfaces/coral_web/src/hooks/experiementalFeatures.ts
misspia-cohere: opting to return entire object

src/interfaces/coral_web/src/hooks/experimentalFeatures.ts
```diff
@@ -0,0 +1,20 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { ExperimentalFeatures, useCohereClient } from '@/cohere-client';
+
+export const useExperimentalFeatures = () => {
```
wujessica, line 5:
```suggestion
/**
 * @description Hook to get experimental features on/off status.
 */
export const useExperimentalFeatures = () => {
```

src/interfaces/coral_web/src/cohere-client/client.ts
```diff
@@ -318,8 +322,32 @@ export class CohereClient {
     return body as Deployment[];
   }
 
+  public async listExperimentalFeatures(): Promise<ExperimentalFeatures> {
```
wujessica, line 325: toiny nit: since this isn't returning us an array i'd hesitate to use the word `list`
```suggestion
  public async getExperimentalFeatures(): Promise<ExperimentalFeatures> {
```

wujessica (approved):
looks good, just 2 nits! can we update the PR screenshot too for posterity? :)

wujessica (approved):

PR #62, opened 2024-04-26
coral-web: support langchain multihop

wujessica:
**Description:**

This PR adds support for rendering tool events sent from the langchain multihop chat endpoint, including adding the functionality to render images from the python interpreter in multihop mode.

**Tests:**

To test, add `USE_EXPERIMENTAL_LANGCHAIN=true` to your `.env` file. Then, select the `python_interpreter` and `internet search` tools and ask a query like `plot the 3 tallest mountains in the world`.

https://github.com/cohere-ai/cohere-toolkit/assets/5898755/bb1b0c5e-0734-4f9c-afe6-cb4f56110e46

### DIFF
--- a/src/interfaces/coral_web/src/cohere-client/client.ts
+++ b/src/interfaces/coral_web/src/cohere-client/client.ts
@@ -176,6 +176,39 @@ export class CohereClient {
     });
   }
 
+  public async langchainChat({
+    request,
+    headers,
+    signal,
+    onOpen,
+    onMessage,
+    onClose,
+    onError,
+  }: {
+    request: CohereChatRequest;
+    headers?: Record<string, string>;
+    signal?: AbortSignal;
+    onOpen?: FetchEventSourceInit['onopen'];
+    onMessage?: FetchEventSourceInit['onmessage'];
+    onClose?: FetchEventSourceInit['onclose'];
+    onError?: FetchEventSourceInit['onerror'];
+  }) {
+    const chatRequest = mapToChatRequest(request);
+    const requestBody = JSON.stringify({
+      ...chatRequest,
+    });
+    return await fetchEventSource(this.getEndpoint('langchain-chat'), {
+      method: 'POST',
+      headers: { ...this.getHeaders(), ...headers },
+      body: requestBody,
+      signal,
+      onopen: onOpen,
+      onmessage: onMessage,
+      onclose: onClose,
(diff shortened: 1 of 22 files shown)


src/backend/main.py
wujessica, line 1: note: any changes to do with `/experimental_features` endpoint will be obsolete once #49 is merged

src/interfaces/coral_web/src/cohere-client/client.ts
```diff
@@ -318,8 +355,33 @@ export class CohereClient {
...
+        response.status
+      );
+    }
+
+    return body as ExperimentalFeatures;
+  }
+
   private getEndpoint(
-    endpoint: 'upload' | 'chat-stream' | 'conversations' | 'tools' | 'deployments'
+    endpoint:
```
misspia-cohere, line 377: I feel it makes more sense to change this to `endpoint: string` at this point 🤔 the list is growing + the method is also private anyway. We're just having to type things out twice

src/interfaces/coral_web/src/hooks/chat.ts
```diff
@@ -248,28 +292,31 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
 
               saveCitations(generationId, citations, documentsMap);
 
+              const finalText = isRAGOn
+                ? replaceTextWithCitations(
+                    // TODO(jessica): temporarily use the text generated from the stream when MAX_TOKENS
```
misspia-cohere, line 297: feel like we should use handles instead of names for this repo 

```suggestion
                    // TODO(@wujessica): temporarily use the text generated from the stream when MAX_TOKENS
```

misspia-cohere (approved):
nicee

src/interfaces/coral_web/src/hooks/chat.ts
wujessica: good call!

src/interfaces/coral_web/src/cohere-client/client.ts
wujessica: though it is repetitive it is nice for the times we do reuse endpoints (e.g. `conversations`) but i don't feel too strongly either way, down to update it next time we add an endpoint!

misspia-cohere (approved):

misspia-cohere (approved):

PR #71, opened 2024-04-29
backend: Add Tool `is_available` and `error_message` fields, tool config cleanup

tianjing-li:
- Removes the `User-Id` header check from the `GET /tools` endpoint - this is not needed
- Modifies existing Tools config to include `is_available` and `error_message` fields. `is_available` is calculated from a `is_available()` method on each Tool class.
- Cleans up imports for tools, adding `__init__.py` files where needed, and including the relevant classes.


lusmoura (approved):
Great PR! 🚀

PR #77, opened 2024-04-30
Deployment: add local model deployment option

lusmoura:
**Description:** adds a local model deployment option using llama cpp and updates the docs.


BeatrixCohere (approved):
Should we add a test here for the final prompt created?

BeatrixCohere (approved):
lgtm

PR #81, opened 2024-04-30
github actions: add build + lint check workflow for coral_web 

misspia-cohere:
Adds build and lint check workflow for `coral_web`

### DIFF
--- a/.github/actions/js-install/actions.yml
+++ b/.github/actions/js-install/actions.yml
@@ -11,7 +11,7 @@ runs:
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
-        node-version-file: "js/.nvmrc"
+        node-version-file: "src/interfaces/coral_web/.nvmrc"
         cache: "pnpm"
         cache-dependency-path: "js/pnpm-lock.yaml"
 
--- a/.github/actions/js-install/actions.yml
+++ b/.github/actions/js-install/actions.yml
@@ -13,9 +13,9 @@ runs:
       with:
         node-version-file: "src/interfaces/coral_web/.nvmrc"
         cache: "pnpm"
-        cache-dependency-path: "js/pnpm-lock.yaml"
+        cache-dependency-path: "src/interfaces/coral_web/pnpm-lock.yaml"
 
     - name: Install dependencies with pnpm
       shell: bash
-      working-directory: js
+      working-directory: src/interfaces/coral_web
       run: pnpm install --frozen-lockfile
--- a/.github/actions/js-install/actions.yml
+++ b/.github/actions/js-install/actions.yml
@@ -1,0 +1,21 @@
+name: js-install
+description: Set up Node.js and pnpm, install dependencies
+
+runs:
+  using: composite
+  steps:
+    - uses: pnpm/action-setup@v2.1.0
+      with:
+        version: "9.0.6"
+
+    - name: Install Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: "js/.nvmrc"
+        cache: "pnpm"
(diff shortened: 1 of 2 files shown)


.github/workflows/interface_tests.yml
```diff
@@ -0,0 +1,23 @@
+name: Interface Tests
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - src/interfaces/coral_web/**
```
misspia-cohere, line 7: [source](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore)

.github/workflows/interface_tests.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: src/interfaces/coral_web
+
+steps:
+  - name: Checkout repo
+    uses: actions/checkout@v3
+  - name: Check coral_web is formatted
+    run: pnpm format
```
wujessica, line 21: before we can use `pnpm` do we need to install it or does it just work :o

misspia-cohere: added, good call 👍

.github/workflows/interface_tests.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: src/interfaces/coral_web
+
+steps:
+  - name: Checkout repo
+    uses: actions/checkout@v3
+    with:
+      fetch-depth: 2
```
misspia-cohere, line 21: leaving this at the default value of `0` fetches the entire history from all branches. Instead, we just fetch the 2 most recent commits to save on time and resources

tianjing-li (approved):
Wow!

.github/actions/js-install/actions.yml
```diff
@@ -0,0 +1,21 @@
...
+  using: composite
+  steps:
+    - uses: pnpm/action-setup@v2.1.0
+      with:
+        version: "9.0.6"
+
+    - name: Install Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: "js/.nvmrc"
```
wujessica, line 14: this should all be `coral_web` instead of `js` right? 😅

misspia-cohere: trueee-- corrected https://github.com/cohere-ai/cohere-toolkit/pull/81/commits/efa9edd8801dbb3b2c861c1d69b56f4effd6089a

.github/actions/js-install/actions.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: pnpm/action-setup@v2.1.0
+      with:
+        version: "9.0.6"
+
+    - name: Install Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: "src/interfaces/coral_web/.nvmrc"
+        cache: "pnpm"
+        cache-dependency-path: "js/pnpm-lock.yaml"
```
wujessica, line 16: i think this also needs to be updated

.github/actions/js-install/actions.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - name: Install dependencies with pnpm
+      shell: bash
+      working-directory: js
```
wujessica, line 20: same with here?

.github/actions/js-install/actions.yml
misspia-cohere: :')

wujessica (approved):
🚀

PR #82, opened 2024-04-30
coral_web: Add is_available check to tools

misspia-cohere:
**Description**
- disables tools that have a false `is_available` value
- add tool `description` as tooltip, else  use `error_message`

Screenshot 2024-05-01 at 11 13 48

### DIFF
--- a/src/interfaces/coral_web/src/components/Configuration/Tools.tsx
+++ b/src/interfaces/coral_web/src/components/Configuration/Tools.tsx
@@ -41,7 +41,7 @@ const ToolSection = () => {
   const { params, setParams } = useParamsStore();
   const { data } = useListTools();
   const { tools: paramTools } = params;
-  const tools = data?.filter((t) => t.is_visible) ?? [];
+  const tools = data?.filter((t) => t.is_visible && t.is_available) ?? [];
   const enabledTools = paramTools ?? [];
   const name = `select-all-tools`;
 
--- a/src/interfaces/coral_web/src/components/Configuration/Tools.tsx
+++ b/src/interfaces/coral_web/src/components/Configuration/Tools.tsx
@@ -41,7 +41,7 @@ const ToolSection = () => {
   const { params, setParams } = useParamsStore();
   const { data } = useListTools();
   const { tools: paramTools } = params;
-  const tools = data?.filter((t) => t.is_visible && t.is_available) ?? [];
+  const tools = data?.filter((t) => t.is_visible) ?? [];
   const enabledTools = paramTools ?? [];
   const name = `select-all-tools`;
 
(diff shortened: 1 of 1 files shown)


src/interfaces/coral_web/src/components/Configuration/Tools.tsx
wujessica, line 1: there's also a part of this task to read and display the `error_message` field!

misspia-cohere: oh I see-- like a tooltip next to the tool maybe?

misspia-cohere: suggesting tooltip instead of rendering directly in the list of tools as some messages can be lengthy

ie.
Screenshot 2024-04-30 at 17 31 47

wujessica: wait do these all usually have errors LOL but they still work? 🤣 confused haha

src/interfaces/coral_web/src/components/Configuration/Tools.tsx
```diff
@@ -95,14 +95,14 @@ const ToolSection = () => {
...
-            {tools.map(({ name }) => {
+            {tools.map(({ name, error_message }) => {
               const enabledTool = enabledTools.find(
                 (enabledTool) => enabledTool.name.toLocaleLowerCase() === name.toLocaleLowerCase()
               );
               const checked = !!enabledTool;
 
               return (
-                <Fragment key={name}>
+                <div key={name} className="flex gap-x-1">
```
wujessica, line 105: let's centre the button!
```suggestion
                <div key={name} className="flex items-center gap-x-1">
```

src/interfaces/coral_web/src/components/Configuration/Tools.tsx
```diff
@@ -95,14 +95,14 @@ const ToolSection = () => {
...
                   <Checkbox
                     checked={checked}
                     onChange={async (e) => {
```
wujessica, line 108: not part of this but I don't think we need this `async`
```suggestion
                    onChange={(e) => {
```

wujessica (approved):

src/interfaces/coral_web/src/components/Configuration/Tools.tsx
misspia-cohere: seems like it ¯\_(ツ)_/¯

wujessica (approved):

tianjing-li:
I think we want to show the ones that have `is_available=False`, but perhaps just make them red (or some other indicator that they're not selectable) and to show their error_message. And the tooltip message should use `description` if it's available, but use `error_message` if its not available. Lmk if that makes sense

tianjing-li (changes requested):
Added some feedback

wujessica (approved):

tianjing-li (approved):
Approved - ty for changing

PR #86, opened 2024-05-02
backend: Set up next.js to proxy requests to the API

malexw:
Google Cloud Run only allows us to have one open port for the application. This means that if we want to run the backend and the frontend in the same container on GCR, the client won't be able to directly reach the backend and we'll need to proxy through the next.js application serving the frontend code.

This change adds a `rewrite` rule to the next.js config to allow for that proxying. The way this PR is written now, all API requests would be changed to route through next.js, though it would be possible to make that configurable to skip a hop.

### DIFF
--- a/src/backend/routers/conversation.py
+++ b/src/backend/routers/conversation.py
@@ -56,7 +56,7 @@ def get_conversation(
     return conversation
 
 
-@router.get("/", response_model=list[ConversationWithoutMessages])
+@router.get("", response_model=list[ConversationWithoutMessages])
 def list_conversations(
     *, offset: int = 0, limit: int = 100, session: DBSessionDep, request: Request
 ) -> list[ConversationWithoutMessages]:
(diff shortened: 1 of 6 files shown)


lusmoura (approved):

tianjing-li (approved):

Justicea83:
These changes seem to have broken the application when it's run locally in docker.

malexw:
@Justicea83 Would you be able to try setting `NEXT_PUBLIC_API_HOSTNAME` in your `.env` file to `http://backend:8000` and tell me if that changes anything for you?

Justicea83:
> @Justicea83 Would you be able to try setting `NEXT_PUBLIC_API_HOSTNAME` in your `.env` file to `http://backend:8000` and tell me if that changes anything for you?

Yes tried and it worked, but you need to update the `run-all` script here: [here](https://github.com/cohere-ai/cohere-toolkit/blob/725fdd9025c58cb522b1de82f130949262aafa82/src/backend/cli/main.py#L45) to make the setup following the readme work

PR #90, opened 2024-05-02
Add baseline Architecture to support auth + User sessions + Basic Auth

tianjing-li:
Beefy first PR to setup plug & play Auth:

- Adds new `auth` folder in services
- Adds `BaseAuthenticationStrategy` abstract class and `BasicAuthentication` (email+password) to start with. The base class should be inherited from all Auth strategies, next steps should include adding OAuth strategies.
- Adds `auth` routes, can retrieve session, logout, login, auth (to implement). The /login method should be generic and run for all auth strategies
- Adds `SessionMiddleware` to track current user session
- Adds extensive unit testing


src/backend/config/auth.py
```diff
@@ -0,0 +1,5 @@
+from backend.services.auth import BasicAuthentication
+
+ENABLED_AUTH_STRATEGIES = [
+    BasicAuthentication,
+]
```
tianjing-li, line 5: The config to set which auth strategies are active

src/backend/main.py
```diff
@@ -15,32 +18,47 @@
...
     app.include_router(chat_router)
     app.include_router(user_router)
     app.include_router(conversation_router)
     app.include_router(tool_router)
     app.include_router(deployment_router)
     app.include_router(experimental_feature_router)
 
+    # Add auth
+    for auth in ENABLED_AUTH_STRATEGIES:
+        if auth.SHOULD_ATTACH_TO_APP:
```
tianjing-li, line 43: This TODO is for setting up OAuth-based strategies, which will eventually require registering them to the FastAPI app during runtime

src/backend/main.py
```diff
@@ -15,32 +18,47 @@
 
 load_dotenv()
 
+ORIGINS = ["*"]
```
rodrigue-h, line 21: let's constrain this before merging - or allow someone to constrain via configuration?

tianjing-li: I just capitalized the const, but we originally had `origins = ["*"]` set to streamline the deployments, without it a lot of the cloud deployments would require an extra step of modifying this in code, which we thought wasn't great UX

src/backend/services/auth/basic.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Returns:
+            str: Hashed password
+        """
+        return bcrypt.hashpw(plain_text_password.encode("utf-8"), bcrypt.gensalt())
+
+    @staticmethod
+    def check_password(plain_text_password: str, hashed_password: str) -> bool:
+        """
+        Hashes a given plain-text password with a randomly generated salt.
```
scott-cohere, line 50: i think this description might be an accidental copy paste from above

src/backend/services/auth/basic.py
```diff
@@ -0,0 +1,89 @@
...
+            plain_text_password (str): Password to check.
+            hashed_password (str): Password to check against.
+
+        Returns:
+            bool: Whether the plain-text password matches the given hashed password.
+        """
+        return bcrypt.checkpw(plain_text_password.encode("utf-8"), hashed_password)
+
+    @classmethod
+    def login(cls, session: Session, payload: dict[str, str]) -> bool:
```
scott-cohere, line 62: return type should be dict | none

scott-cohere (approved):
just two tiny things. good stuff! 😃

src/backend/services/auth/basic.py
tianjing-li: ty

scott-cohere (approved):

scott-cohere (approved):

PR #94, opened 2024-05-03
tools: Update default NEXT_PUBLIC_API_HOSTNAME for the new api routing

malexw:
With the new routing of API requests through next.js, we need to update the default `NEXT_PUBLIC_API_HOSTNAME` value set in the `.env` file when running setup.

### DIFF
--- a/src/backend/cli/main.py
+++ b/src/backend/cli/main.py
@@ -42,7 +42,7 @@ WELCOME_MESSAGE = r"""
 """
 DATABASE_URL_DEFAULT = "postgresql+psycopg2://postgres:postgres@db:5432"
 PYTHON_INTERPRETER_URL_DEFAULT = "http://localhost:8080"
-NEXT_PUBLIC_API_HOSTNAME_DEFAULT = "http://localhost:8000"
+NEXT_PUBLIC_API_HOSTNAME_DEFAULT = "http://backend:8000"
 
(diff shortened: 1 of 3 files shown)


src/backend/cli/main.py
```diff
@@ -42,7 +42,7 @@ class ToolName(StrEnum):
 """
 DATABASE_URL_DEFAULT = "postgresql+psycopg2://postgres:postgres@db:5432"
 PYTHON_INTERPRETER_URL_DEFAULT = "http://localhost:8080"
-NEXT_PUBLIC_API_HOSTNAME_DEFAULT = "http://localhost:8000"
+NEXT_PUBLIC_API_HOSTNAME_DEFAULT = "http://backend:8000"
```
lusmoura, line 45: We define this variable in multiple places, such as .env_template, README, and some FE files. Can you please update all the occurrences?

malexw: I _think_ I've got it now in all the places where we need it. Outside of the containers, `localhost:8000` is still the right way to reach the backend (e.g. when using the Postman collections). We just want to make sure that the frontend container is using `backend:8000` when we're running locally with `docker compose`.

lusmoura (approved):

lusmoura:
Merging to unblock another user that's having the same issue

PR #95, opened 2024-05-03
Removed setup only dependency for makefile

samuelzxu:

- [ ] **PR title**: "area: environment setup"

- [ ] **PR message**:  I've tried to install for debian 12 using the linux friendly instructions in the quickstart and the option 2 installation instructions. However, the `make setup` or `make first-run` commands fail with dependency errors. I think this is because of the `--only setup` restriction, which could restrict the dependency installs from pyproject.toml solely to setup dependencies rather than all dependencies
    - **Description:** Removes the setup-only restriction
    - **Dependencies:** none

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ reset-db:
 	docker compose down
 	docker volume rm cohere_toolkit_db
 setup:
-	poetry install --only setup --verbose
+	poetry install --verbose
 	poetry run python3 src/backend/cli/main.py
 lint:
 	poetry run black .


tianjing-li:
Can I get more context on the issue? The initial `poetry install` with `--only setup` installs those dependencies on your local host machine to run the `main.py` CLI script, however `make migrate` and `make dev` should build the docker images for you (which runs a `poetry install` for all dependencies).

samuelzxu:
The problem that I had was that the `make setup` script didn't run. It resulted in transformers and llama_cpp dependency errors. After removing the `--only setup` flag, the script run without errors.

tianjing-li:
Hey @samuelzxu, the Makefile has been updated since the creation of this PR, I'll close this for now but if you're running into issues with the current make commands feel free to creating a new issue/PR as needed, thanks!

PR #100, opened 2024-05-03
Add Elasticsearch log retriever

Justicea83:

- [ ] **PR title**: "Add search to ELK cluster"
  - Add changes to add search from ELK stack
  - Example: "deployment: Cohere Platform"

    - **Description:** Add Elasticsearch retriever
    - **Dependencies:** `elasticsearch`

Screenshot 2024-05-03 at 19 02 05

Screenshot 2024-05-03 at 19 17 45


.env-template
```diff
@@ -27,4 +27,8 @@ AZURE_CHAT_ENDPOINT_URL=<ENDPOINT URL>
 USE_EXPERIMENTAL_LANGCHAIN=False
 
 # Community features
-USE_COMMUNITY_FEATURES='True'
\ No newline at end of file
+USE_COMMUNITY_FEATURES='True'
+
+# searching elastic search cluster
+ELASTICSEARCH_HOST=
```
tianjing-li, line 33: Nit: Can you move this under the `#TOOLS` section? And change the comment to `#Searching Elasticsearch cluster`

src/backend/config/tools.py
```diff
@@ -12,6 +12,7 @@
     LangChainVectorDBRetriever,
     LangChainWikiRetriever,
     TavilyInternetSearch,
```
tianjing-li, line 14: We recently added a new `backend/community/tools` folder, can you move all logic there instead?

src/backend/config/tools.py
```diff
@@ -32,6 +33,7 @@ class ToolName(StrEnum):
     Python_Interpreter = "Python_Interpreter"
     Calculator = "Calculator"
     Tavily_Internet_Search = "Internet Search"
+    Logs_Retriever = "App Logs"
```
tianjing-li, line 36: Rename this to clarify that the source is Elasticsearch

src/backend/config/tools.py
```diff
@@ -95,6 +97,20 @@ class ToolName(StrEnum):
...
+        name=ToolName.Logs_Retriever,
+        implementation=LogsRetriever,
+        kwargs={
+            "host_url": os.getenv("ELASTICSEARCH_HOST", None),
+            "index": os.getenv("ELASTICSEARCH_INDEX", None),
+            "deployment_name": ToolName.Logs_Retriever
+        },
+        is_visible=True,
+        is_available=LogsRetriever.is_available(),
+        error_message="LogsRetriever not available.",
```
tianjing-li, line 110: Add here that the host and index env variables are required

src/backend/tools/retrieval/logs_retriever.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(
+            self,
+            host_url: str,
+            index: str,
+            deployment_name: str,
+    ) -> None:
+        self.index = index
+        self.deployment_name = deployment_name
+        self.client = Elasticsearch(host_url)
+        self.cohere_api_key = os.environ.get("COHERE_API_KEY")
```
tianjing-li, line 21: It looks `cohere_api_key`  is not required?

src/backend/tools/retrieval/nyc_service_request_retriever.py
```diff
@@ -0,0 +1,105 @@
...
+
+from backend.chat.custom.model_deployments.deployment import get_deployment
+from backend.schemas.cohere_chat import CohereChatRequest
+from backend.services.logger import get_logger
+
+cohere_api_key = os.environ.get("COHERE_API_KEY")
+logger = get_logger()
+
+
+class NYCServiceRequestRetriever(BaseRetrieval):
```
tianjing-li, line 16: It looks like this retriever tool isn't present in the dict mapping, should we add it? Or was this for testing?

src/backend/tools/retrieval/logs_retriever.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class LogsRetriever(BaseRetrieval):
+    def __init__(
+            self,
+            host_url: str,
+            index: str,
+            deployment_name: str,
+    ) -> None:
+        self.index = index
```
tianjing-li, line 18: Suggestion: Rather than instantiating `LogsRetriever()` with the index and host, can you add these as params on the class, eg :

```
class LogsRetriever(BaseRetrieval):
    index = os.environ.get(.., None)
    host = os.environ.get(.., None)
```

Then your `__init__()` method would use `self.index` and `self.host`,

And your classmethod 
`is_available()` should check `return all[self.index is not None, self.host is not None]`

tianjing-li (changes requested):
Requested some changes

PR #109, opened 2024-05-07
backend: Refactor model deployments folder

tianjing-li:
- Refactors the `model_deployments` folder to move it into the parent `backend` folder
- Changes `models` to `database_models`


src/community/model_deployments/hugging_face.py
```diff
@@ -1,11 +1,10 @@
 import logging
-from typing import Any, Dict, Generator, List
+from typing import Any, Dict, List
 
-from cohere.types import NonStreamedChatResponse, StreamedChatResponse
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
+from backend.model_deployments.base import BaseDeployment
```
lusmoura, line 6: Should we import this from community as it was before?

src/community/model_deployments/local_model.py
```diff
@@ -1,10 +1,9 @@
-import logging
 from typing import Any, Dict, List
 
 from llama_cpp import Llama
 
+from backend.model_deployments.base import BaseDeployment
```
lusmoura, line 5: Same as above

lusmoura (approved):

src/community/model_deployments/hugging_face.py
tianjing-li: done

src/community/model_deployments/local_model.py
tianjing-li: done

PR #111, opened 2024-05-07
Routers: add versioning

lusmoura:
Add v1 to all existing endpoints and fix FE routes


tianjing-li:
What do you think about adding a `v1` folder inside `router` that will contain all the current routes?

lusmoura:
> What do you think about adding a `v1` folder inside `router` that will contain all the current routes?

I thought about it, but I didn't want to overcomplicate the current structure. I'd suggest adding a new folder once we add a v2. What do you think? I can do it now if you have a strong opinion about adding the folder tho

tianjing-li:
@lusmoura nah no big deal, we probably won't have a v2 for a while, I think this is good to go

tianjing-li (approved):
good to go

scott-cohere (approved):

PR #127, opened 2024-05-09
chore: increase codeowners for contributors

GangGreenTemperTatum:

- [X] **PR title**: "Increases `CODEOWNERS` for `cohere-toolkit` team"


scott-cohere (approved):

scott-cohere:
@GangGreenTemperTatum looks like this is blocked because of unsigned commits, could you sign them? or if its easier i can just make the same PR

PR #128, opened 2024-05-09
chore: increase codeowners signed

GangGreenTemperTatum:
superseeds #127


scott-cohere (approved):
thanks ads!

scott-cohere:
@GangGreenTemperTatum hmm still seems not be signed. i just created a PR on your behalf https://github.com/cohere-ai/cohere-toolkit/pull/130, could you ptal?

PR #136, opened 2024-05-14
Deployments: add bedrock

lusmoura:
Add Bedrock model deployment, update documentation and CLI


src/backend/config/deployments.py
```diff
@@ -51,6 +53,17 @@ class ModelDeploymentName(StrEnum):
             "AZURE_CHAT_ENDPOINT_URL",
         ],
     ),
+    ModelDeploymentName.Bedrock: Deployment(
+        name=ModelDeploymentName.Bedrock,
+        deployment_class=BedrockDeployment,
+        models=BedrockDeployment.list_models(),
+        is_available=BedrockDeployment.is_available(),
+        env_vars=[
+            "BEDROCK_PROFILE_NAME",
```
tianjing-li, line 62: Do we need the access key here?

lusmoura: Good catch!

tianjing-li (approved):
Looks good!

PR #138, opened 2024-05-15
Auth: Add Google OAuth

tianjing-li:
- Adds base Google OAuth logic
- Modifies existing /login and /auth endpoints to interface with these Auth strategies

Note: OAuth not in usable state after merging this, will require JWT token

TODO: 
- Remove SessionMiddleware usage, replace with JWT token auth
- E2E testing with FE


src/backend/routers/auth.py
```diff
@@ -33,57 +51,102 @@ def get_session(request: Request):
...
+    return RedirectResponse(url="/")
+
+
+@router.post("/auth")
+async def authenticate(request: Request, auth: Auth, session: DBSessionDep):
+    """
+    Authentication endpoint used for OAuth strategies. Logs the user in the redirect environment and then
+    sets the current session with the user returned from the auth token.
+
+    Args:er
```
malexw, line 112: Is the `:er` part a typo?

malexw (approved):
There's nothing that really jumps out at me here, though it should probably get a review from someone who's more familiar with the backend than me.

src/backend/routers/auth.py
```diff
@@ -1,14 +1,32 @@
...
+
+    Returns:
+        List[str]: List of names for enabled Authentication strategies, can be empty.
+    """
+    return ENABLED_AUTH_STRATEGY_MAPPING.keys()
+
+
 @router.get("/session")
 def get_session(request: Request):
     """
```
EugeneLightsOn, line 32: Maybe it would be better to rename this method too. Also, I'm not completely clear on why we need session middleware for the REST API.

.env-template
```diff
@@ -29,5 +29,9 @@ USE_EXPERIMENTAL_LANGCHAIN=False
 # Community features
 USE_COMMUNITY_FEATURES='True'
 
-# Auth session
+# Authentication session
 SESSION_SECRET_KEY=<GENERATE_A_SECRET_KEY>
+
+# Google OAuth
+GOOGLE_CLIENT_ID=<GOOGLE_CLIENT_ID>
+GOOGLE_CLIENT_SECRET=<GOOGLE_CLIENT_SECRET>
```
EugeneLightsOn, line 37: We need to clarify how to obtain the values of these variables

tianjing-li:
@EugeneLightsOn I've added this ticket: https://linear.app/cohereai/issue/EXT2-74/add-jwt-token-bearer-header-to-requests-remove-usage-of-session 

To track the work of removing session and using a JWT token based auth

malexw (approved):
🚢 🇮🇹

PR #144, opened 2024-05-17
infra: Add a code coverage % report via Check Runs API

malexw:
As part of ensuring continuing code quality, we'd like to add a check that rejects PRs if the test coverage drops too low. Part 1 of that change is just to report existing coverage on new PRs, which this change implements.


BeatrixCohere (approved):
One quick comment

tianjing-li (approved):
lgtm - thanks for this

PR #151, opened 2024-05-22
Set model deployment configuration through the UI at runtime (verse .env)

BeatrixCohere:
- Updates Coral to set the deployment options in the request verse .env file 
- Return available deployment at one deployment call verse multiple 
- Update sagemaker auth configuration 
- clean up env vars in backend

**AI Description**

This PR introduces changes to the backend and frontend services. 

## Backend

The PR adds a new `utils.py` module in the `src/backend/model_deployments` directory, which defines the `get_model_config_var` function. This function retrieves the model config variable value based on the provided variable name and model config. 

The `get_deployment` function, previously defined in `src/backend/model_deployments/utils.py`, has been moved to `src/backend/chat/custom/utils.py`. This function now takes an additional `model_config` parameter and returns a `BaseDeployment` instance based on the provided deployment name and model configuration. 

The `ModelDeploymentName` class in `src/backend/config/deployments.py` has been updated to import environment variables from the corresponding model deployment modules (`BEDROCK_ENV_VARS`, `COHERE_ENV_VARS`, `AZURE_ENV_VARS`, and `SAGE_MAKER_ENV_VARS`). 

The `is_available` method in the `AzureDeployment`, `BedrockDeployment`, `CohereDeployment`, and `SageMakerDeployment` classes has been modified to check if all required environment variables are set instead of checking specific variable values. 

The `get_deployment` function in `src/backend/chat/custom/custom.py` now passes the `model_config` parameter to the `get_deployment` function. 

The `chat` method in the `CustomChat` class in `src/backend/chat/custom/custom.py` has been updated to include the `model_config` parameter in the function signature. 

## Frontend

The PR removes the line `# For SageMaker: The line below for AWS configure file to sync credentials` and `$HOME/.aws:/root/.aws` from the `docker-compose.yml` file. 

## Summary

To summarise, the PR introduces changes to the backend services by adding a new `utils.py` module, updating the `get_deployment` function, modifying the `ModelDeploymentName` class and related deployment classes, and updating the `get_deployment` and `chat` methods. Additionally, the PR removes specific lines related to SageMaker configuration from the `docker-compose.yml` file in the frontend services.


src/interfaces/coral_web/src/components/EditEnvVariablesButton.tsx
wujessica, line 1: We can probably clean this up a bit now that it's simpler! i.e.
- update the submit button to say `Save` 
- remove all the functionality to do with the error message since it can't error

src/interfaces/coral_web/src/components/EditEnvVariablesButton.tsx
```diff
@@ -37,17 +37,29 @@ export const EditEnvVariablesButton: React.FC<{ className?: string }> = () => {
 };
 
 /**
- * @description Renders a modal to edit a selected deployment's .env variables.
+ * @description Renders a modal to edit a selected deployment's config
  */
-const EditEnvVariablesModal: React.FC<{ onClose: () => void }> = ({ onClose }) => {
+export const EditEnvVariablesModal: React.FC<{
+  onClose: () => void;
+  defaultDeployment: string;
```
wujessica, line 44:
```suggestion
  defaultDeployment: string;
  onClose: () => void;
```

src/interfaces/coral_web/src/components/EditEnvVariablesButton.tsx
```diff
@@ -92,7 +104,11 @@ const EditEnvVariablesModal: React.FC<{ onClose: () => void }> = ({ onClose }) =
 
     try {
       setIsSubmitting(true);
-      await updateDeploymentEnvVariables({ name: deployment, env_vars: envVariables });
```
wujessica, line 95: does this mean we don't need this endpoint? could we delete it :o

wujessica (approved):
frontend side looks good!

src/backend/routers/chat.py
```diff
@@ -172,6 +178,12 @@ def process_chat(
     """
     user_id = request.headers.get("User-Id", "")
     deployment_name = request.headers.get("Deployment-Name", "")
+    model_config = {}
+    if not request.headers.get("Deployment-Config", "") == "":
+        model_config = dict(
```
tianjing-li, line 183: It's unclear to me what `Deployment-Config` input should be, and what it becomes when assigned to `model_config`, would recommend adding some comments here to clarify

tianjing-li: If I understand the code, it looks like we want to add a config dictionary we can pass to request headers to define a model deployment, which we can then call during the request (without having to preregister variables in the .env), is my understanding correct?

BeatrixCohere: Yes that is correct, so you can set model config per request

src/interfaces/coral_web/src/components/EditEnvVariablesButton.tsx
BeatrixCohere: We will most likely need it in the future.

PR #155, opened 2024-05-24
tools: Update default API hostname to localhost:8000

malexw:
When we removed the proxy from the frontend, we didn't update the CLI setup script to use `localhost:8000` as the default backend instead of `backend:8000`.

This pull request updates the `NEXT_PUBLIC_API_HOSTNAME_DEFAULT` in the `main.py` file.

## Summary
The pull request updates the `NEXT_PUBLIC_API_HOSTNAME_DEFAULT` variable in the `main.py` file to use "http://localhost:8000" as the default API hostname instead of "http://backend:8000".

## Changes
- Updates `NEXT_PUBLIC_API_HOSTNAME_DEFAULT` to "http://localhost:8000".


wujessica (approved):

tianjing-li (approved):
Good catch ty!

PR #157, opened 2024-05-27
backend: Make poetry groups optional by default

tianjing-li:
- Without the optional flag set here, the previous `poetry install` would actually install all the groups. This limits them, moves some dependencies to more appropriate groups, and ultimately will help limit the docker image size


BeatrixCohere (approved):
lgtm

lusmoura (approved):

PR #163, opened 2024-05-31
backend (auth): Remove Session usage, add JWT bearer auth logic, dynamically set router dependencies

tianjing-li:
- Removes previous `SessionMiddleware` usage, and `get_session` endpoint
- Fixes the `GET /auth_strategies` endpoint
- Modifies `/login` and `/auth` to return a JWT token on successful authentication
- Adds `config/routers.py` file to define default or auth dependencies. These dependencies are set during app creation time
- Adds unit tests


src/backend/tests/routers/test_chat.py
```diff
@@ -101,6 +101,10 @@ def test_fail_chat_missing_user_id(
         headers={"Deployment-Name": ModelDeploymentName.CoherePlatform},
     )
 
+    import pdb
```
lusmoura, line 104: Looks like we should remove this

lusmoura (approved):
LGTM! What should we do about the tests that are failing? Should we add `JWT_SECRET_KEY` to the repo?

EugeneLightsOn (approved):
Looks good to me, thanks. Only take a look at Luisa's comments.

src/backend/tests/routers/test_chat.py
tianjing-li: definitely, ty

tianjing-li: For testing, I'll disable auth for now ,I'll have to revisit it later

src/backend/services/auth/request_validators.py
```diff
@@ -0,0 +1,39 @@
+from fastapi import HTTPException, Request
+
+from backend.services.auth.jwt import JWTService
+
+
+def validate_authorization(request: Request):
```
lusmoura, line 6: nit: a few functions don't have the return type

src/backend/config/routers.py
```diff
@@ -0,0 +1,88 @@
...
+    CHAT = "chat"
+    CONVERSATION = "conversation"
+    DEPLOYMENT = "deployment"
+    EXPERIMENTAL_FEATURES = "experimental_features"
+    TOOL = "tool"
+    USER = "user"
+
+
+# Router dependency mappings
+ROUTER_DEPENDENCIES = {
```
lusmoura, line 24: Can we add some documentation about this? Or maybe tests? Just something to remind people to add dependencies whenever adding a new router.

src/backend/services/auth/request_validators.py
tianjing-li: done

src/backend/config/routers.py
tianjing-li: Sure, adding some more documentation

Makefile
```diff
@@ -22,17 +22,10 @@ reset-db:
 setup:
 	poetry install --with setup,community --verbose
 	poetry run python3 src/backend/cli/main.py
-win-setup:
```
EugeneLightsOn, line 25: Why do we need to remove it? Windows deployments depend on it

Makefile
```diff
@@ -22,17 +22,10 @@ reset-db:
...
-	poetry install --with setup,community --verbose
-	poetry run python src/backend/cli/main.py
 lint:
 	poetry run black .
 	poetry run isort .
 first-run:
 	make setup
 	make migrate
 	make dev
-win-first-run:
```
EugeneLightsOn, line 35: Why do we need to remove it? Windows deployments depend on it.

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
     build:
       context: .
       args:
-        INSTALL_COMMUNITY_DEPS: false
+        INSTALL_COMMUNITY_DEPS: true
```
malexw, line 33: Just double-checking that we want to change this

EugeneLightsOn (changes requested):
Great PR, thank you 
Please restore Make win-setup and win-first-run commands and good to go.

malexw (approved):
I didn't take a close look at the tests, but this looks pretty good to me

Makefile
tianjing-li: Good catch

docker-compose.yml
tianjing-li: Good catch

EugeneLightsOn (approved):
LGTM, thanks

PR #164, opened 2024-06-03
Tools: unify retrievers/functions and add file tools

lusmoura:
This is the first of a series of PRs to introduce multihop to the toolkit.
- Removes the duplicate logic for retrievers and function tools from custom chat, since all retrievers are tools for multihop.
- Adds both `read_file` and `search_document` tools to support uploaded files and tools simultaneously.
- The tools mentioned above required some changes to the tool calls, such as passing the model deployment (to rerank the chunks) and the user id (to query the DB for files).
- The tools also required a change to the `chat_history` to add an utterance containing the available files before generating the tool calls.

Looking for review for chunking documents and manipulating `chat_history`.

**AI Description**

This PR introduces two new tools, `ReadFileTool` and `SearchFileTool`, which allow for reading and searching within uploaded files. 

The changes include:
- Addition of `ReadFileTool` and `SearchFileTool` classes in `src/backend/tools/files.py`.
- Updates to `src/backend/chat/custom/custom.py` to handle the new tools and add file content to the chat history.
- Updates to `src/backend/config/tools.py` to include the new tools and adjust existing tool definitions.
- Addition of `get_files_by_file_names` function in `src/backend/crud/file.py` to retrieve files by their names.
- Addition of `file_content` field in `src/backend/database_models/file.py` to store file content.
- Updates to `src/backend/model_deployments/base.py`, `src/backend/model_deployments/azure.py`, `src/backend/model_deployments/bedrock.py`, and `src/backend/model_deployments/cohere_platform.py` to include `chat_history` in the `invoke_tools` method.
- Updates to `src/backend/routers/conversation.py` to read file content during file upload.
- Addition of `SYSTEM` role in `src/backend/schemas/chat.py` and `src/backend/schemas/cohere_chat.py` to identify system messages.
- Addition of `tool_results` field in `src/backend/schemas/cohere_chat.py` to store results from invoking tools.
- Addition of imports for `ReadFileTool` and `SearchFileTool` in `src/backend/tools/__init__.py`.
- Update to `poetry.lock` to include the new dependency `pypdf`.


src/backend/chat/custom/custom.py
```diff
@@ -39,116 +40,68 @@ def chat(self, chat_request: CohereChatRequest, **kwargs: Any) -> Any:
...
+        tool_names = [tool.name for tool in tools]
+        if ToolName.Read_File or ToolName.SearchFile in tool_names:
+            chat_history = self.add_files_to_chat_history(
+                chat_history,
+                conversation_id,
+                kwargs.get("session"),
+                kwargs.get("user_id"),
+            )
+
+        tools_to_use = deployment_model.invoke_tools(
```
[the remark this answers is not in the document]
lusmoura: Do you mean the `directly_answer` tool or an actual answer?

src/backend/tools/files.py
```diff
@@ -0,0 +1,146 @@
...
+                files_dicts.append(
+                    {
+                        "text": chunk,
+                        "title": file.file_name,
+                        "url": file.file_path,
+                    }
+                )
+
+        # Combine and rerank the documents
+        result = combine_documents({query: files_dicts}, model_deployment)
```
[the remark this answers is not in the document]
lusmoura: Like each tool should return only one result and we should chunk, combine and rerank once we have documents for every tool?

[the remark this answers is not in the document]
BeatrixCohere: Yes, but each tool returns multiple outputs/documents and we should chunk, combine and rerank them

src/backend/chat/custom/custom.py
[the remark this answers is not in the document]
BeatrixCohere: An actual answer

src/backend/tools/files.py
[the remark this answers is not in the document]
lusmoura: Since we're moving away from using documents and are now using tool_restuls, what should we use to rerank the outputs? Some tools might not have a query 🤔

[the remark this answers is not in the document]
BeatrixCohere: Oh yes, we should only rerank on internet search / ones with a query - but chunk on them all to my knowledge

src/backend/chat/custom/custom.py
```diff
@@ -39,116 +40,68 @@ def chat(self, chat_request: CohereChatRequest, **kwargs: Any) -> Any:
...
+        tools = [
+            Tool(**AVAILABLE_TOOLS.get(tool.name).model_dump())
+            for tool in chat_request.tools
+            if AVAILABLE_TOOLS.get(tool.name)
+        ]
 
-        Args:
-            file_paths (list[str]): File paths.
-            req_tools (list[str]): Required tools.
+        if len(tools) == 0:
```
tianjing-li, line 61: Nit: prefer the slightly more pythonic `if not tools:`

src/backend/chat/collate.py
```diff
@@ -3,78 +3,118 @@
...
-        for x in zip_longest(*documents.values(), fillvalue=None)
-        for y in x
-        if y is not None
-    ]
+        # Merges the results with the same tool call and parameters
+        tool_call_hashable = str(tool_call)
+        if tool_call_hashable not in reranked_results.keys():
+            reranked_results[tool_call_hashable] = {"call": tool_call, "outputs": []}
+        
+        reranked_results[tool_call_hashable]["outputs"].extend([chunked_outputs[r.index] for r in res.results if r.relevance_score > RELEVANCE_THRESHOLD])
```
lusmoura, line 79: Had to add a threshold because the model would take too long to generate responses when passing too many tool results

src/backend/chat/custom/custom.py
```diff
@@ -38,133 +41,188 @@ def chat(self, chat_request: CohereChatRequest, **kwargs: Any) -> Any:
...
+        # If the second event is a tool call, the tools are invoked, and the results are returned
+        # Otherwise, the chat response is returned as a direct answer
+        stream_start_event = next(stream)
+
+        if stream_start_event is None:
+            yield [], False
+
+        second_event = next(stream)
+
+        if second_event["event_type"] == StreamEvent.TOOL_CALLS_GENERATION:
```
BeatrixCohere, line 168: It could also be a tool calls partial generation, I think it is better to loop thorough the whole stream and when we see tool calls generation then call the code below

src/backend/chat/collate.py
```diff
@@ -3,78 +3,126 @@
...
-        # Only rerank on text of document
-        # TODO handle no text in document
-        docs_to_rerank = [doc["text"] for doc in documents]
+    reranked_results = {}
+    for tool_result in tool_resuls:
+        tool_call = tool_result["call"]
 
-        # If no documents to rerank, continue to the next query
-        if not docs_to_rerank:
+        if not tool_call.parameters.get("query") and not tool_call.parameters.get(
```
BeatrixCohere, line 50: Can we add comments here?

src/backend/chat/collate.py
```diff
@@ -3,78 +3,126 @@
...
-        all_rerank_docs[query] = [documents[r.index] for r in res.results]
+        query = tool_call.parameters.get("query") or tool_call.parameters.get(
+            "search_query"
+        )
 
-    return all_rerank_docs
+        chunked_outputs = []
+        for output in tool_result["outputs"]:
+            text = output.get("text")
+            if not text:
```
BeatrixCohere, line 62: We should still include output that doesn't have "text"

src/backend/chat/collate.py
```diff
@@ -3,78 +3,126 @@
...
-        # TODO handle no text in document
-        docs_to_rerank = [doc["text"] for doc in documents]
+    reranked_results = {}
+    for tool_result in tool_resuls:
+        tool_call = tool_result["call"]
 
-        # If no documents to rerank, continue to the next query
-        if not docs_to_rerank:
+        if not tool_call.parameters.get("query") and not tool_call.parameters.get(
+            "search_query"
```
BeatrixCohere, line 51: We should still add the outputs to be called

src/backend/tools/files.py
```diff
@@ -0,0 +1,95 @@
+from typing import Any, Dict, List
+
+from pypdf import PdfReader
```
BeatrixCohere, line 3: Should we use the langchain tools?

src/backend/chat/custom/custom.py
```diff
@@ -31,140 +34,194 @@ def chat(self, chat_request: CohereChatRequest, **kwargs: Any) -> Any:
...
 
-                chat_request.tools = None
-                if kwargs.get("stream", True) is True:
-                    return deployment_model.invoke_chat_stream(
-                        chat_request,
-                        tool_results=tool_results,
-                    )
+            for event, generated_direct_answer in stream:
+                if generated_direct_answer:
+                    should_call_chat = False
```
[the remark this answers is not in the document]
lusmoura: I couldn't get it to work with the whole generator thing, but I can try again

src/backend/tools/files.py
lusmoura: Yeah, we were using the langchain version of the pdf reader before, but I was not sure if we should keep using their packages or the actual module. Wdyt?

src/backend/chat/custom/custom.py
lusmoura: can we get partial generation with single step?

BeatrixCohere (approved):

src/backend/chat/collate.py
```diff
@@ -3,78 +3,130 @@
 
 from backend.model_deployments.base import BaseDeployment
 
+RELEVANCE_THRESHOLD = 0.5
+
 
 def combine_documents(
```
scott-cohere, line 9: this function just calls and returns `rerank_and_chunk` now. seems redundant to wrap it, maybe we can just call `rerank_and_chunk` directly now?

src/backend/chat/collate.py
```diff
@@ -3,78 +3,130 @@
...
             continue
 
-        res = model.invoke_rerank(query=query, documents=docs_to_rerank)
-        # Sort the results by relevance score
-        res.results.sort(key=lambda x: x.relevance_score, reverse=True)
-        # Map the results back to the original documents
-        all_rerank_docs[query] = [documents[r.index] for r in res.results]
+        query = tool_call.parameters.get("query") or tool_call.parameters.get(
+            "search_query"
+        )
```
scott-cohere, line 69: im assuming it is either `search_query` or `query` because the param defs in our tools, is there a problem with just keeping them consistent? eg. make them all `query` and just check for `query`?

scott-cohere (approved):
lgtm, just a few nits

src/backend/chat/collate.py
```diff
@@ -3,78 +3,130 @@
...
+        # Sort the results by relevance score
+        res.results.sort(key=lambda x: x.relevance_score, reverse=True)
+
+        # Map the results back to the original documents
+        # Merges the results with the same tool call and parameters
+        tool_call_hashable = str(tool_call)
+        if tool_call_hashable not in reranked_results.keys():
+            reranked_results[tool_call_hashable] = {"call": tool_call, "outputs": []}
+
+        reranked_results[tool_call_hashable]["outputs"].extend(
```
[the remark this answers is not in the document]
lusmoura: I changed the code so that we merge all outputs for a tool_call before reranking them. It should fix this issue.

BeatrixCohere (approved):

scott-cohere (approved):

PR #170, opened 2024-06-04
[backend] Remove community dependancies from make first run

BeatrixCohere:
Remove the community dependencies from make first run 

**AI Description**

This pull request updates the `Makefile` to adjust the dependencies installed during the setup process.

## Summary
The PR removes the `community` option from the `poetry install` command, reducing the number of dependencies installed during the setup process.

## Changes
- The `poetry install` command in the `setup` and `win-setup` targets no longer includes the `community` option.


Makefile
```diff
@@ -20,10 +20,10 @@ reset-db:
...
 setup:
-	poetry install --with setup,community --verbose
+	poetry install --with setup --verbose
 	poetry run python3 src/backend/cli/main.py
 win-setup:
-	poetry install --with setup,community --verbose
+	poetry install --with setup --verbose
 	poetry run python src/backend/cli/main.py
 lint:
 	poetry run black .
```
tianjing-li, line 29: In `cli/main.py` there's some logic revolving `community_tools_prompt()` that might need to be removed since it would break now

tianjing-li (approved):

PR #172, opened 2024-06-04
coral-web: add icon font back into codebase

wujessica:
**Description:**
This PR adds the icon font back into the codebase so that hosted version of the app may have the proper icons in place in case CORs has not yet been updated. We also usually need to update `icons.css` when we add more icons so it makes sense to have the icons locally.

**AI Description**


src/interfaces/coral_web/src/styles/fonts.css
```diff
@@ -1,6 +1,6 @@
 @font-face {
   font-family: 'CohereText';
-  src: url('https://fonts.cohere.com/CohereVariable.woff2') format('woff2'),
+  src: url('https://fonts.cohere.com/CohereText-Regular.woff2') format('woff2'),
```
EugeneLightsOn, line 3: Just to clarify, do we need to move this font to /public/fonts too, or does it work well as is?

EugeneLightsOn (approved):
Looks good to me, just one small clarification

src/interfaces/coral_web/src/styles/fonts.css
wujessica: We can't move the fonts back into the codebase because they're not open source unfortunately. :( But at least if these fail we can fallback on a system font vs for icons there's no fallback

PR #182, opened 2024-06-07
Remove enforced auth token for /user endpoints

tianjing-li:
- Removes the validation for Authorization: Bearer for user endpoints to allow creating users


malexw (approved):
I notice that with this change, endpoint for listing all users, getting user details, and deleting users are all unprotected. I assume that this is just temporary and we'll be adding more detailed permissions around these with the addition of organizations?

malexw (approved):

PR #184, opened 2024-06-07
Chat: update response

lusmoura:
With the new version of the Cohere SDK, chat_history can contain tool results, which was breaking existing tests

**AI Description**

The PR introduces a new `TOOL` role for `ChatRole` in the `Chat` module. This change affects both the schema definition in `chat.py` and the `generate_chat_response` function in `services/chat.py`.

**Files changed:**
- `src/backend/schemas/chat.py`

**Code changes:**
- Added a new `TOOL` role to the `ChatRole` `StrEnum` in `chat.py`.
- Modified the `ChatMessage` class in `chat.py` to include an optional `tool_results` field, which stores a list of tool call results.
- Updated the default value of the `tool_results` field in `chat.py` to an empty list.
- Refactored the `generate_chat_response` function in `services/chat.py` to handle the new `tool_results` field in the `ChatMessage` class.

**Summary:**
The PR introduces a new `TOOL` role for the `ChatRole` enumeration, enhancing the flexibility of the chat system by allowing messages to be associated with tools. This change is reflected in both the schema definition and the chat response generation logic. The `tool_results` field in the `ChatMessage` class provides a way to store and retrieve results from tool calls, improving the chat system's ability to handle tool-related interactions and responses.

### DIFF
--- a/src/backend/services/chat.py
+++ b/src/backend/services/chat.py
@@ -588,7 +588,6 @@ def generate_chat_response(
         documents=documents,
         search_results=response.get("search_results", []),
         event_type=StreamEvent.NON_STREAMED_CHAT_RESPONSE,
-        is_finished=True,
         conversation_id=conversation_id,
         tool_calls=tool_calls,
     )
@@ -625,7 +624,6 @@ def generate_langchain_chat_stream(
             ChatResponseEvent(
                 event=StreamEvent.STREAM_START,
                 data=StreamStart(
-                    is_finished=False,
                     conversation_id=conversation_id,
                 ),
             )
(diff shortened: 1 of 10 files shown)


src/backend/services/chat.py
```diff
@@ -549,8 +549,9 @@ def generate_chat_response(
 
         chat_history.append(
             ChatMessage(
-                role=message["role"],
-                message=message["message"],
+                role=message.get("role", ""),
```
tianjing-li, line 552: Should we default to one of chatbot/user/system here?

tianjing-li (approved):

PR #185, opened 2024-06-07
Metrics: add middleware

lusmoura:
**AI Description**

This PR introduces a new `MetricsMiddleware` class to the backend, enabling the reporting of metrics such as request duration and status codes.

**Code Changes:**
- Added `MetricsMiddleware` class in `src/backend/services/metrics.py`.
- Imported necessary modules: `os`, `time`, `requests`, `BaseHTTPMiddleware`, and `Request`.
- Defined `REPORT_ENDPOINT` using `os.getenv`.
- Implemented `dispatch` and `report_metrics` methods in `MetricsMiddleware` to record request duration.
- Added error handling for potential exceptions during metric reporting.
- Included helper methods `get_endpoint_name` and `get_status_code` to extract endpoint name and status code, respectively.
- Integrated `MetricsMiddleware` in `src/backend/main.py` by importing it and adding it to the middleware stack.


src/backend/routers/chat.py
```diff
@@ -69,6 +73,7 @@ async def chat_stream(
                 session=session,
                 conversation_id=conversation_id,
                 user_id=user_id,
+                trace_id=trace_id,
```
BeatrixCohere, line 76: Should we have a session type object that has all the session info including the DBsession above?

lusmoura: I added the minimum for now since Tanzim is OOO. Once he's back, I'll catch up with him and check what else we should add.

src/backend/services/metrics.py
```diff
@@ -0,0 +1,161 @@
...
+
+        return data
+
+    def get_method(self, scope):
+        try:
+            return scope["method"].lower()
+        except KeyError:
+            return "unknown"
+        except Exception as e:
+            logging.warning("Failed to get method:", e)
```
scott-cohere, line 66: nit, we use fstrings everywhere else to format our warning logs. lets do the same here?

src/backend/routers/agent.py
```diff
@@ -40,6 +40,7 @@ def create_agent(session: DBSessionDep, agent: CreateAgent, request: Request) ->
         tools=agent.tools,
     )
 
+    request.state.agent = agent_data
```
scott-cohere, line 43: What is this for?

src/backend/model_deployments/sagemaker.py
```diff
@@ -73,6 +78,17 @@ def is_available(cls) -> bool:
...
+        # TODO: implement metrics correctly
+        start_time = time.perf_counter()
+        metrics_data = MetricsData(
+            endpoint_name="co.chat",
+            method="POST",
+            trace_id=kwargs.pop("trace_id", None),
+            user_id_hash=hash_string(kwargs.pop("user_id", None)),
+            model=chat_request.model,
+            success=True,
+        )
```
scott-cohere, line 90: so we're metering the route on toolkit which calls this and is also metered? (if im understanding correctly?) is this what we want to do, maybe now or later we can just use traces.

same to all the other calls we're doing in the other deployments, etc.

src/backend/routers/agent.py
lusmoura: When creating an agent we need to return it's data, like name, description...

scott-cohere (approved):

src/backend/chat/collate.py
```diff
@@ -21,6 +21,9 @@ def rerank_and_chunk(
     Returns:
         List[Dict[str, Any]]: List of reranked and combined documents.
     """
+    trace_id = kwargs.get("trace_id", "")
+    user_id = kwargs.get("user_id")
```
tianjing-li, line 25: Should this have a default of "" as well, or should we deal with an edge case where user_id might be None?

src/backend/chat/custom/custom.py
```diff
@@ -117,6 +126,8 @@ def is_not_direct_answer(self, event: Dict[str, Any]) -> bool:
         )
 
     def call_chat(self, chat_request, deployment_model, **kwargs: Any):
+        trace_id = kwargs.get("trace_id")
+        user_id = kwargs.get("user_id")
```
tianjing-li, line 130: Same here for both trace_id and user_id, should they have defaults?

src/backend/chat/custom/custom.py
```diff
@@ -230,6 +243,8 @@ def get_managed_tools(self, chat_request: CohereChatRequest):
         ]
 
     def get_tool_calls(self, tools, chat_history, deployment_model, **kwargs: Any):
+        trace_id = kwargs.get("trace_id")
+        user_id = kwargs.get("user_id")
```
tianjing-li, line 247: Same

src/backend/model_deployments/azure.py
```diff
@@ -53,47 +60,109 @@ def is_available(cls) -> bool:
         return all([os.environ.get(var) is not None for var in AZURE_ENV_VARS])
 
     def invoke_chat(self, chat_request: CohereChatRequest, **kwargs: Any) -> Any:
-        yield self.client.chat(
-            **chat_request.model_dump(exclude={"stream"}),
-            **kwargs,
+        start_time = time.perf_counter()
```
tianjing-li, line 63: Same comment for azure/bedrock/cohere/sagemaker, but I think the metrics logic should be refactored out of the actual Model deployment implementation for 2 reasons: DRY and separation of concerns. 

Suggestion:

You can implement some sort of decorator function that takes as arg the Model Deployment chat method (eg, `CoherePlatform.invoke_chat`), and implements the logic of creating the MetricsData, try/except/finally etc. 

Then the idea would be that you could have the metrics in a separate file somewhere (perhaps services?)

And implement your decorator like: 

```
import time
from functools import wraps
from typing import Any, Callable, Dict

def collect_metrics(func: Callable) -> Callable:
    @wraps(func)
    def wrapper(self, chat_request: 'CohereChatRequest', **kwargs: Any) -> Any:
             #.. do metrics stuff
             response = {}
             try:
                      response = func(self, chat_request, **kwargs)
                      response_dict = to_dict(response)
             except..
             finally.. etc
```

Then your methods could just be decorated with this 

```
@collect_metrics
def invoke_chat(etc):
```

tianjing-li: Would help clean up the code a lot!

tianjing-li: The `report_metrics` code also looks fine to me, you would probably just move it to the same file `collect_metrics` live, or put them into a class (although I'm not too familiar with encapsulating decorators in a class, the syntax might be different)

tianjing-li (changes requested):
Couple changes

src/backend/model_deployments/azure.py
lusmoura: Great suggestion, the code looks much better now. I kept the decorators specific to each function because they have some particularities and I thought it would be confusing to keep everything together (i.e. streaming and non streaming responses)

tianjing-li (approved):
Much cleaner! Approved

PR #175, opened 2024-06-05
Agent DB Migration

scott-cohere:
Thank you for contributing to the Cohere Toolkit!

- [ ] **PR title**: "area: description"
  - Where "area" is whichever of interface, frontend, model, tools, backend, etc. is being modified. Use "docs: ..." for purely docs changes, "infra: ..." for CI changes.
  - Example: "deployment: add Azure model option"

- [ ] **PR message**: ***Delete this entire checklist*** and replace with
    - **Description:** a description of the change
    - **Issue:** the issue # it fixes, if applicable
    - **Dependencies:** any dependencies required for this change

- [ ] **Add tests and docs**: Please include testing and documentation for your changes
- [ ] **Lint and test**: Run `make lint` and `make run-tests`

**AI Description**

PR #186, opened 2024-06-07
[backend] agents API

scott-cohere:

- [x] **PR title**: "area: description"

This PR builds on top of https://github.com/cohere-ai/cohere-toolkit/pull/175. In #175 we set up the DB and agents entity model. In this PR we set up the new API routes the logic write unit tests for the API logic and added postman examples.

**AI Description**

This PR introduces the ability to manage agents, including creating, listing, updating, and deleting them. It also adds tests for these new functionalities.

**Code Changes:**
- Added new router for agents in src/backend/routers/agent.py.
- Added agent-related functions in src/backend/routers/agent.py.
- Updated src/backend/config/routers.py to include `AGENT` in `RouterName and added `RouterName.AGENT` configuration.
- Added `ModelType` and `DeploymentType` enums in src/backend/schemas/agent.py.
- Updated Agent, CreateAgent, UpdateAgent, and added DeleteAgent in src/backend/schemas/agent.py.
- Updated update_agent function in src/backend/crud/agent.py to exclude `None` values when updating agent attributes.
- Updated Agent class in src/backend/database_models/agent.py to make `version`, `description`, `preamble`, and `temperature` nullable.
- Added tests for agent creation, listing, updating, and deletion in src/backend/tests/routers/test_agent.py.
- Added Alembic migration script in src/backend/alembic/versions/1cbb45f9fe25_.py to update the `agents` table.


src/backend/routers/agent.py
```diff
@@ -1,3 +1,144 @@
...
+
+router = APIRouter(
+    prefix="/v1/agents",
+)
+router.name = RouterName.AGENT
+
+
+@router.post("", response_model=Agent)
+def create_agent(session: DBSessionDep, agent: CreateAgent, request: Request):
+    user_id = get_header_user_id(request)
```
tianjing-li, line 18: Looks like we don't use user_id anywhere, so you actually don't need to call `get_header_user_id`

tianjing-li: Same thing for all endpoints

docs/postman/Toolkit.postman_collection.json
```diff
@@ -423,6 +423,158 @@
 				}
 			]
 		},
+		{
+			"name": "Agents",
```
lusmoura, line 427: Nice!

src/backend/database_models/agent.py
```diff
@@ -23,11 +23,11 @@ class Model(StrEnum):
 class Agent(Base):
     __tablename__ = "agents"
 
-    version: Mapped[int] = mapped_column(Integer, default=1)
+    version: Mapped[int] = mapped_column(Integer, default=1, nullable=True)
```
lusmoura, line 26: In which case would this be nullable?

lusmoura (approved):

src/backend/database_models/agent.py
scott-cohere: i think i misunderstood initially. i changed it so that none of these columns should take in nullable values and if nothing is passed in they should default to the default values here.

the only required values should be `userid`, `model`, `deployment`, `name`. which is reflected by a lack of default value and ive written the unit tests in this PR to check these missing cases.

lusmoura (approved):

src/backend/routers/agent.py
tianjing-li: Yup that works!

tianjing-li (approved):

PR #193, opened 2024-06-11
perf: downgrade rehype-highlight and generate once blob url

tomtobac:
## Description

- Move the blob generation to `STREAM_END` event and in chat history

- Pin version `rehype-highligh@6.0.0`(1) to avoid a memory leak which "looks"(2) like it landed in hightligh.js until they fix it.

[1] [Comparison](https://github.com/rehypejs/rehype-highlight/compare/6.0.0...7.0.0) of version 6 to 7

[2] Memory leak [thread](https://github.com/remarkjs/react-markdown/issues/791) in the `react-markdown`


src/interfaces/coral_web/package.json
```diff
@@ -89,4 +89,4 @@
     "prettier": "2.8.7",
     "prettier-plugin-tailwindcss": "0.4.1"
   }
-}
+}
```
knajjars, line 92: probably need to run `pnpm format:write`

knajjars (approved):

src/interfaces/coral_web/src/pages/c/[id]/index.tsx
```diff
@@ -40,12 +41,19 @@ const ConversationPage: NextPage<Props> = () => {
...
-  } = useConversation({ conversationId: urlConversationId });
+  const { data, isLoading, isError, error } = useConversation({
+    conversationId: urlConversationId,
+  });
+  const conversation = useMemo(() => {
+    if (!data) return;
+
+    data.messages = data.messages.map((message) => ({
+      ...message,
+      text: replaceCodeBlockWithIframe(message.text),
```
Comment (wujessica) [inline src/interfaces/coral_web/src/pages/c/[id]/index.tsx:52]: would it make sense to move this to `mapHistoryToMessages`? since we do mapping for `replaceTextWithCitations` there too

wujessica (approved):

wujessica (approved):

PR #195, opened 2024-06-11
fix: citations being removed on last stream chunk

knajjars:
This PR fixes a bug recently introduced where we used `originalText` instead of `text` which contains citations to render the message content.

**AI Description**

This pull request updates the `MessageContent` component in the `src/interfaces/coral_web/src/components/MessageContent.tsx` file. Specifically, it modifies how the component handles displaying code blocks within messages.

## Changes
- Updates the `md` variable assignment when handling fulfilled messages. Instead of using `message.originalText`, it now uses `message.text` to replace the code block with an iframe.


wujessica (approved):
🙏

PR #202, opened 2024-06-12
feat: Add Blacklist logic on logout

tianjing-li:
- Adds Blacklist logic
- Adds JWT token_id to Blacklist table when calling /logout
- Checks Blacklist table wherever `Authorization: Bearer` header is checked


src/backend/routers/auth.py
```diff
@@ -133,17 +136,23 @@ async def oidc_authenticate(request: Request, session: DBSessionDep):
 
 
 @router.get("/logout", response_model=Logout)
-async def logout(request: Request):
+async def logout(
+    request: Request,
+    session: DBSessionDep,
+    token: dict | None = Depends(validate_authorization),
```
scott-cohere, line 142: what does having `Depends(validate_authorization)` as the default value here do? Also `depends` isn't imported here, is the line by accident?

src/backend/routers/auth.py
```diff
@@ -133,17 +136,23 @@ async def oidc_authenticate(request: Request, session: DBSessionDep):
...
 
     Args:
         request (Request): current Request object.
 
     Returns:
         dict: Empty on success
     """
-    # TODO: Design blacklist
+    if token is not None:
+        db_blacklist = Blacklist(token_id=token["jti"])
```
scott-cohere, line 154: what is the `token` query param defined in this router function suppose to be? i see nothing is passed in there with the unit test https://github.com/cohere-ai/cohere-toolkit/pull/202/files#diff-2ea68412ab67e98ac5c98863b626382b19b598be7396e5db65c96c17d69f882cR93. not sure if im misunderstanding fastapi syntax here 😅

src/backend/routers/auth.py
tianjing-li: When you add any `Depends()` calls, whether during router definition or endpoint definition FastAPI will call those dependencies (in the order you added them) on that request. So here I'm adding the `validation_authorization` dependency. 

And nice catch, added the `Depends` back, think I forgot it during merge conflict

src/backend/routers/auth.py
tianjing-li: The `validate_authorization` dependency returns a decoded token in this case

tianjing-li: It returns a dict or None, so we still need to validate here `if token is not None:`, then adds it to the blacklist

PR #204, opened 2024-06-12
[backend] adding tools for agents, filter tools by agent_id

scott-cohere:
Eventually we would want to map tools to a DB table. However, for v1 we can just store tools as a list of tool names as users cannot actually delete,add,update tools in the application itself (must be done in the code). For v1 we also do not have the notion of different organizations yet.

Also adding query param on `GET /v1/tools` -> `GET /v1/tools?agent_id=...` to filter tool by agent id

**AI Description**

This PR introduces a new `tools` field to the `Agent` model, allowing for a list of associated tools. This change is reflected in the database schema, test suites, and API endpoints.

- The `Agent` model in src/backend/database_models/agent.py now includes a `tools` field, which is a list of `ToolName` enums.
- The upgrade and downgrade functions in src/backend/alembic/versions/922e874930bf_.py have been updated to add and remove the `tools` column from the `agents` table, respectively.
- The create_agent function in src/backend/routers/agent.py now includes the tools parameter, which is passed to the `create_agent` function in the `agent_crud` module.
- The list_tools API endpoint in src/backend/routers/tool.py has been modified to accept an optional `agent_id` parameter. It now returns a list of tools associated with the specified agent, if provided.
- The Agent class in src/backend/schemas/agent.py now includes a tools field, which is a list of `ToolName`.
- The test_create_agent and test_update_agent functions in src/backend/tests/crud/test_agent.py have been updated to include the `tools` field in the agent creation and update requests.
- The Agent factory in src/backend/tests/factories/agent.py now includes a tools field, which is a list of random `ToolName` enums.
- The test_create_agent function in src/backend/tests/routers/test_agent.py has been updated to include the `tools` field in the agent creation request.
- The test_list_tools function in src/backend/tests/routers/test_tool.py has been modified to test the listing of tools associated with a specific agent.


lusmoura (approved):
LGTM!

PR #206, opened 2024-06-13
Chat: add multistep

lusmoura:
- Adds logic for multistep:
   - For each step, try to generate tool calls: 
       - if there are tool calls, call them and try again
       - if there are no more tool calls, return
- Refactors handle chat stream
- Updates non streaming chat so that behind the scenes it uses the same streaming chat
- Adds new stream events, such as tool-calls-chunk

### Streaming chat with multiple steps
image

### Streaming chat with one step
image

### Streaming chat with custom tool
image

### Streaming chat with force single step
image

### Non streaming chat with multiple steps
image

 **AI Description**

The PR updates the Cohere Python SDK to the latest version, `5.5.7`. This includes changes to the `custom.py` file, which introduces a new `CustomChat` class that extends `BaseChat`. 

Here are the changes in detail:

- **poetry.lock**: Updates the `cohere` package version to `5.5.7`.
- **pyproject.toml**: Updates the `cohere` dependency to `^5.5.7`.
   - Adds a `MAX_STEPS = 15` constant.
   - Modifies the `chat` method of the `CustomChat` class:
     - Adds `self.chat_request` and `self.is_first_start` attributes.
     - Introduces a `for` loop that iterates `MAX_STEPS` times, invoking `self.call_chat` and logging the step.
     - Removes the `handle_managed_tools` and `get_tool_results` methods.
   - Adds the `is_final_event` method, which checks if an event is final based on the presence of tool calls and managed tools.
   - Adds the `handle_event` method, which handles stream start, stream end, and other events differently.
   - Adds the `is_not_direct_answer` method, which checks if an event contains tool calls.
   - Adds the `call_chat` method, which invokes the chat stream and handles the results.
   - Adds the `get_managed_tools` method, which returns a list of tools based on the chat request's tools.
   - Adds the `get_tool_calls` method, which invokes the chat stream and handles tool calls and chat history updates.
   - Adds the `handle_tool_calls_stream` method, which processes the tool calls stream and returns the chat history, a copy of the stream, and a flag indicating a direct answer.
   - Removes the `add_files_to_chat_history` method.
- **src/backend/chat/collate.py**: Updates the `rerank_and_chunk` function to use `tool_call.get("parameters")` instead of `tool_call.parameters`.
- **src/backend/schemas/chat.py**: 
   - Adds a `tool_calls` field to the `ChatMessage` class.
   - Adds a `text` field to the `StreamEnd` class.
   - Adds a new `StreamToolCallsChunk` class with `tool_call_delta` and `text` fields.
   - Updates the `StreamEventType` union to include the new `StreamToolCallsChunk` class.
- **src/backend/schemas/cohere_chat.py**: Adds a `force_single_step` field to the `CohereChatRequest` class.
- **src/backend/schemas/tool.py**: Adds a new `ToolCallDelta` class with `name`, `index`, and `parameters` fields.
   - Adds imports for `logging` and `ToolCallDelta`.
   - Updates the `generate_chat_response` function to include `force_single_step` in the `CohereChatRequest` model dump.
   - Adds the `generate_chat_response` function, which generates a chat response from a model deployment stream.
   - Updates the `generate_chat_stream` function to use the new `generate_chat_response` function.
   - Removes the `generate_chat_response` function.
   - Adds the `handle_stream_event` function, which handles different stream event types and dispatches them to specific handlers.
   - Adds helper functions for handling different stream event types: `handle_stream_start`, `handle_stream_text_generation`, `handle_stream_search_results`, `handle_stream_search_queries_generation`, `handle_stream_tool_calls_generation`, `handle_stream_citation_generation`, `handle_stream_tool_calls_chunk`, and `handle_stream_end`.
- **src/backend/model_deployments/base.py**: Removes the `invoke_tools` method from the `BaseDeployment` class.
- **src/backend/model_deployments/azure.py** and **src/backend/model_deployments/cohere_platform.py**: Updates the `invoke_tools` method to use `invoke_chat_stream` instead of directly invoking the chat stream.


src/backend/services/chat.py
```diff
@@ -368,6 +371,63 @@ def update_conversation_after_turn(
...
+        model_deployment_stream,
+        response_message,
+        conversation_id,
+        user_id,
+        should_store,
+        **kwargs,
+    )
+
+    non_streamed_chat_response = None
+    for event in stream:
```
scott-cohere, line 409: This is because non-streaming API doesn't support multistep, right? Do we plan to support non-streaming multistep in API later? If so, maybe worth adding comment/todo here about that.

src/backend/services/chat.py
```diff
@@ -508,103 +490,214 @@ def generate_chat_stream(
...
         update_conversation_after_turn(
-            session, response_message, conversation_id, final_message_text, user_id
+            session, response_message, conversation_id, stream_end_data["text"], user_id
         )
 
 
-def generate_chat_response(
-    session: DBSessionDep,
-    model_deployment_response: Generator[StreamedChatResponse, None, None],
+def handle_stream_event(
```
scott-cohere, line 497: great refactor!

scott-cohere (approved):

src/backend/services/chat.py
lusmoura: It actually does! Instead of implementing multistep for streaming and nonstreaming, I added only streaming multistep, and the nonstreaming endpoint simply returns the last event without any of the intermediate steps.

lusmoura: Will add a comment to clarify

PR #207, opened 2024-06-13
[backend] track conversation and assistant relationship + update openAPI client

scott-cohere:
- Add new `agent_id` column to conversation table. FKey to an ID in agent table.
- Update `list_conversations` endpoint to allow query parameter to filter by agent_id
- Unit test
- Update openAPI client


src/backend/crud/conversation.py
```diff
@@ -37,13 +37,21 @@ def get_conversation(
     """
     return (
         db.query(Conversation)
-        .filter(Conversation.id == conversation_id, Conversation.user_id == user_id)
+        .filter(
```
tianjing-li, line 40: Can we add the conditional filter here instead? You can chain .filter() operations 

eg:
```
query = <the previous query>
if agent_id is not None:
      query = query.filter(Conversation.agent_id == agent_id)

return query.first()
```

src/backend/routers/conversation.py
```diff
@@ -59,21 +59,33 @@ async def get_conversation(
...
+        agent_id (str): Query parameter for agent ID to optionally filter conversations by agent.
         session (DBSessionDep): Database session.
         request (Request): Request object.
 
     Returns:
         list[ConversationWithoutMessages]: List of conversations.
     """
     user_id = get_header_user_id(request)
+
+    if agent_id:
```
tianjing-li, line 84: Now don't need to conditional here, just pass `agent_id` directly and the crud method will handle it

src/backend/database_models/conversation.py
```diff
@@ -18,9 +19,15 @@ class Conversation(Base):
...
+    )
 
     @property
     def messages(self):
         return sorted(self.text_messages, key=lambda x: x.position)
 
-    __table_args__ = (Index("conversation_user_id", user_id),)
+    __table_args__ = (
+        Index("conversation_user_id", user_id),
+        Index("conversation_agent_id", agent_id),
```
tianjing-li, line 32: If conversations are filtered by `agent_id` only while filtering by `user_id`, you could get more benefits by adding a composite index of both `user_id + agent_id`

tianjing-li (approved):
Looks good. just a small comment

PR #230, opened 2024-06-18
chore: add wrench and share icon

knajjars:
Adds Wrench and Share icons

**AI Description**

This pull request adds two new icons, `wrench` and `share`, to the existing set of icons.

- The `Icons` array in `Icon.tsx` has been updated to include the new icon names.
- The `icons.css` file has been modified to define the content for the new icon classes: `.icon-wrench` and `.icon-share`.


src/interfaces/coral_web/src/components/Shared/Icon/Icon.tsx
```diff
@@ -77,6 +77,8 @@ export const Icons = [
   'warning',
   'web',
   'compass',
+  'wrench',
+  'share',
```
wujessica, line 81: i think we already have `link`?

wujessica (approved):
could we update the title to be `area: description` so `coral-web: add wrench and share icon`?

src/interfaces/coral_web/src/components/Shared/Icon/Icon.tsx
wujessica: jk i guess this is `share` 
image

PR #237, opened 2024-06-18
[backend] passing in agent when calling streaming chat

scott-cohere:
Once you create and agent you can pass in the agent_id in the chat-stream endpoint via query parameter.

This PR:
- In chat req validator, check that agent exists
- In chat req validator, if conversation is also passed in with agent chat stream, check that the conversation belongs to that agent
- Unit tests

**AI Description**

The PR introduces changes to the backend of a chatbot application, specifically enhancing the chat functionality. 

**Files Changed:**
- `src/backend/config/routers.py`
- `src/backend/routers/chat.py`
- `src/backend/services/request_validators.py`
- `src/backend/tests/crud/test_conversation.py`
- `src/backend/tests/routers/test_chat.py`

**Changes Include:**
- Introduction of a new parameter, `agent_id: str | None = None`, in the `chat_stream` and `chat` functions in `chat.py`.
- Modification of the `process_chat` function in `chat.py` to include the `agent_id` parameter.
- Updates to the `process_chat` function in `services/chat.py` to include `agent_id` and perform additional checks and settings updates.
- Addition of imports and updates to the `get_or_create_conversation` function in `services/chat.py` to include the `agent_id` parameter.
- Changes to the `validate_chat_request` function in `request_validators.py` to include a `session` parameter and additional validation logic for `agent_id`.
- New tests added and updated in `test_conversation.py` and `test_chat.py` to cover the `agent_id` parameter and related functionality.


src/backend/routers/chat.py
```diff
@@ -108,7 +112,7 @@ async def chat(
         should_store,
         managed_tools,
         deployment_config,
-    ) = process_chat(session, chat_request, request)
+    ) = process_chat(session, chat_request, request, agent_id)
```
scott-cohere, line 115: I think for now FE will just have to get the deployment of the agent and pass it into the header.

Starting to realize this def will need to be a new endpoint, but I can follow up on that after.

lusmoura (approved):

PR #246, opened 2024-06-19
[backend] enforce agent update with user-id

scott-cohere:
Only allow creator of agent to update the agent.

in this PR:
- Update some HTTP codes to be more accurate to the actual error
- Update minor string formatting for errors to be more consistent
- Update the validator for update agent request to cross check the user ID is the creator of the agent
- Update unit tests

**AI Description**

This pull request makes changes to the backend code, specifically related to agent management and validation. 

**Summary:**
- Updates the `status_code` from `404` to `400` for more accurate error handling when an agent is not found.
- Enhances the `validate_update_agent_request` function to include additional checks for the agent ID and user ID, ensuring requests are properly validated.
- Adjusts test cases to reflect the updated status codes and error messages.

**Code Changes:**
- **src/backend/routers/agent.py:**
  - Modified the `status_code` in the `HTTPException` raised when an agent is not found to `400` for more accurate error handling.
- **src/backend/services/request_validators.py:**
  - Added `session: DBSessionDep` as a parameter to the `validate_update_agent_request` function for improved validation.
  - Introduced additional checks to ensure the presence of an agent ID and that the agent belongs to the user making the request.
  - Updated the `status_code` in the `HTTPException` raised when the agent is not found to `400`.
- **src/backend/tests/routers/test_agent.py:**
  - Adjusted the expected status codes and error messages in test cases to reflect the updates made in the code.
  - Added a new test case, `test_update_agent_wrong_user`, to verify that an unauthorized user cannot update an agent.

### DIFF
--- a/src/backend/services/request_validators.py
+++ b/src/backend/services/request_validators.py
@@ -22,7 +22,7 @@ def validate_user_header(request: Request):
 
     """
 
-    user_id = request.headers.get(request)
+    user_id = request.headers.get("User-Id")
     if not user_id:
         raise HTTPException(
             status_code=401, detail="User-Id required in request headers."
--- a/src/backend/services/request_validators.py
+++ b/src/backend/services/request_validators.py
@@ -164,7 +164,7 @@ async def validate_create_agent_request(session: DBSessionDep, request: Request)
         )
 
 
-async def validate_update_agent_request(request: Request):
+async def validate_update_agent_request(session: DBSessionDep, request: Request):
     """
     Validate that the update agent request has valid tools, deployments, and compatible models.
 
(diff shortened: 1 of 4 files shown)


src/backend/services/request_validators.py
```diff
@@ -174,8 +174,22 @@ async def validate_update_agent_request(request: Request):
...
+    if not agent_id:
+        raise HTTPException(status_code=400, detail="Agent ID is required.")
+
+    agent = agent_crud.get_agent_by_id(session, agent_id)
+    if not agent:
+        raise HTTPException(
+            status_code=400, detail=f"Agent with ID {agent_id} not found."
+        )
 
+    if agent.user_id != request.headers.get("User-Id"):
```
tianjing-li, line 187: This should probably use the `get_header_user_id` method so it works with auth

tianjing-li (changes requested):
Small change required

src/backend/services/request_validators.py
scott-cohere: good call out. also changing it for the `validate_user_header` validator as well.

tianjing-li (approved):
lgtm one small change

src/backend/services/request_validators.py
tianjing-li: let's change it back for `validate_user_header`

PR #247, opened 2024-06-19
Propagate should_store logic to the File upload logic

EugeneLightsOn:
**AI Description**

The `process_chat` function in `chat.py` has been modified to include an additional check before attaching files to messages. 

- Previously, the code directly called the `attach_files_to_messages` function to attach files to the chat session. 
- Now, an `if should_store:` condition has been added to control whether files are attached to messages. 

This change allows for more flexibility in handling file attachments, potentially based on specific conditions or user preferences.


scott-cohere (approved):
nice edge case find!

PR #269, opened 2024-06-24
Add basic structure for GDrive OAuth tool 

BeatrixCohere:
Add basic structure for GDrive OAuth tool 

Doesn't include:
- More error handling
- Testing
- storing (user_id,tool_id) in state
- correct UI flow

**AI Description**

## Summary
This PR adds a new tool to the backend, Google Drive, and its associated authentication. It also updates the tool authentication system to include tool-based authentication.

## Changes
- Added a new tool, Google Drive, which searches Google Drive for relevant files.
- Added tool-based authentication, which is currently experimental and in development.
- Added a new table, `tool_auth`, to store access tokens for tools that require authentication.
- Added a new column, `is_auth_required`, to the `ManagedTool` model to indicate whether a tool requires authentication.
- Added a new column, `auth_url`, to the `ManagedTool` model to store the authentication URL for a tool.
- Updated the `list_tools` function to include the `is_auth_required` and `auth_url` fields for each tool.
- Added a new abstract base class, `BaseAuth`, for tool authentication.
- Added new functions, `get_auth_url`, `is_auth_required`, and `process_auth_token`, to the `BaseAuth` class for tool authentication.
- Added a new class, `GoogleDriveAuth`, which inherits from `BaseAuth` and implements authentication for the Google Drive tool.


src/backend/tools/google_drive.py
```diff
@@ -0,0 +1,162 @@
...
+logger = get_logger()
+
+
+class GoogleDrive(BaseTool):
+    """
+    Experimental (In development): Tool that searches Google Drive
+    """
+
+    @classmethod
+    def is_available(cls) -> bool:
```
lusmoura, line 28: Any chance we can check if the user is authenticated here?

lusmoura (approved):
LGTM!

src/backend/tools/google_drive.py
BeatrixCohere: We could but I think it is better to have the error in is_auth_required?

giannis2two:
Should we add a `tool_id` property in `BaseTool` for all tools to share?

BeatrixCohere:
> tool_id

Agreed, won't do here but will follow up with it

PR #284, opened 2024-06-25
coral-web: load tools from agent api

wujessica:
**Description:**
Use `tools` given by the `/agent` api to show tools specific to that agent. Also removes the ability to deselect tools from the composer if we are using the non-base agent.

https://github.com/cohere-ai/cohere-toolkit/assets/5898755/d3ceec49-6135-4fe1-bfe7-093deb506f20

**AI Description**

This PR introduces changes to the settings drawer and tools tab in the user interface. It also updates the `CohereClient` class to use the new `ManagedTool` type.

- The `SettingsDrawer` component now uses the `useAgent` hook to fetch the agent data and dynamically render the tools tab based on the available tools.
- The `ToolsTab` component now accepts an optional `requiredTools` prop, allowing for more flexible rendering of available and unavailable tools.
- The `ComposerToolbar` and `EnabledDataSources` components have been updated to include a `canDisableDataSources` prop, providing more control over data source disabling.
- The `ToggleCard` component now has an optional `disabled` prop, with a default value of `false`.
- The `useChat` hook has been modified to handle `requestTools` and correctly format the tools for the chat request.
- The `ManagedTool` type has been added to the `CohereClient` class, and the `listTools` method now returns a promise of `ManagedTool[]` instead of `Tool[]`].
- The `Composer` component now includes a `canDisableDataSources` prop, allowing for dynamic disabling of data sources.
- The `AgentsPage` component now uses the `useListTools` hook to fetch the available tools and sets the agent tools as parameters.

### DIFF
--- a/src/interfaces/coral_web/src/components/Shared/Markdown/Markdown.tsx
+++ b/src/interfaces/coral_web/src/components/Shared/Markdown/Markdown.tsx
@@ -93,9 +93,11 @@ export const Markdown = ({
   // @ts-ignore
   const components: Components = useMemo(
     () => ({
+      // @ts-ignore
       pre: (props) => <Pre {...props} />,
       p: P,
       references: References,
+      // @ts-ignore
       iframe: Iframe,
       ...customComponents,
     }),
--- a/src/interfaces/coral_web/src/cohere-client/client.ts
+++ b/src/interfaces/coral_web/src/cohere-client/client.ts
@@ -12,6 +12,7 @@ import {
   FinishReason,
   ListAuthStrategy,
   ListFile,
+  ManagedTool,
   Tool,
   UpdateAgent,
   UpdateConversation,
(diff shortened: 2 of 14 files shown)


src/interfaces/coral_web/src/components/Shared/Markdown/Markdown.tsx
```diff
@@ -93,9 +93,11 @@ export const Markdown = ({
   // @ts-ignore
   const components: Components = useMemo(
     () => ({
+      // @ts-ignore
       pre: (props) => <Pre {...props} />,
       p: P,
       references: References,
+      // @ts-ignore
```
wujessica, line 100: needed to fix `pnpm build`

misspia-cohere (approved):
nice! 

should we also add to this description when there are no tools for the agent you're on... something like `${agent name} does not use any tools` 
Screenshot 2024-06-25 at 14 41 48

wujessica:
> nice!
> 
> should we also add to this description when there are no tools for the agent you're on... something like `${agent name} does not use any tools` Screenshot 2024-06-25 at 14 41 48

yea that or we just hide it, maybe we can ask design?

PR #287, opened 2024-06-25
[backend] webscrape tool and scraping during websearch

scott-cohere:
**Web Scrape Demo**

https://github.com/cohere-ai/cohere-toolkit/assets/146760070/b95c0320-bfc6-4974-86dd-09279197051b

**Web Search w/ Scraping Demo**

https://github.com/cohere-ai/cohere-toolkit/assets/146760070/a58729ec-7421-46da-bacd-1eb861c10f80

**AI Description**

This PR introduces a new `WebScrapeTool` tool for web scraping, adds a new exception handler, and makes several other changes and improvements to the backend.

## Changes
- **New `WebScrapeTool` Tool**:
   - Added a new WebScrapeTool class in `src/backend/tools/web_scrape.py` for web scraping.
   - Updated `src/backend/config/tools.py` and `src/backend/tools/__init__.py` to include the new tool.
   - Added tests for the new tool in `src/backend/tests/crud/test_file.py`.
- **Exception Handling**:
   - Added a new exception handler in `src/backend/main.py` to handle global errors.
   - Added a new test client fixture in `src/backend/tests/routers/conftest.py` to test the global error handler.
   - Added a new test in `src/backend/tests/routers/test_main.py` to verify the exception handler.
- **File Handling**:
   - Refactored file handling code in `src/backend/services/file.py` to use a new `get_file_content` function and removed the `FileService` class.
   - Updated `src/backend/routers/conversation.py` to use the new `get_file_content` function and removed `FileService` imports.
   - Added a new `validate_file_size` function in `src/backend/services/file.py` to validate file sizes before uploading.
   - Added a new `get_files_by_user_id` function in `src/backend/crud/file.py` to retrieve files by user ID.
- **Dependency Updates**:
   - Updated the Poetry lock file (`poetry.lock`) with dependency updates and version changes.
   - Added `beautifulsoup4` and `openpyxl` to the list of dependencies in `pyproject.toml`.
- **Other Changes**:
   - Removed `asyncio` and `asynccontextmanager` imports in `src/backend/main.py`.
   - Added `get_logger` import and initialized the `logger` variable in `src/backend/main.py`.
   - Added `Request` import and updated the `startup_event` function in `src/backend/main.py`.
   - Updated the `call_tools` function in `src/backend/chat/custom/custom.py` to log tool results.
   - Added `Depends(get_session)` to the default middleware list in `src/backend/config/routers.py`.


src/backend/tools/tavily.py
```diff
@@ -12,26 +14,79 @@ class TavilyInternetSearch(BaseTool):
...
 
     @classmethod
     def is_available(cls) -> bool:
         return cls.tavily_api_key is not None
 
     def call(self, parameters: dict, **kwargs: Any) -> List[Dict[str, Any]]:
         query = parameters.get("query", "")
-        content = self.client.search(query=query, search_depth="advanced")
+        result = self.client.search(
+            query=query, search_depth="advanced", include_raw_content=True
```
scott-cohere, line 26: scraping can be done with `include_raw_content` which is tad simpler to implement. alternatively I tried it with beautifulsoup which works just as well and provides a bit more flexibility on how the page is parsed. ultimately decided this will suffice.

src/backend/tools/tavily.py
```diff
@@ -12,26 +14,79 @@ class TavilyInternetSearch(BaseTool):
...
-            {
-                "url": result["url"],
-                "text": result["content"],
-            }
-            for result in content["results"]
+            {"url": result["url"], "text": result["content"]}
+            for result in reranked_results
         ]
 
+    def rerank_page_snippets(
```
scott-cohere, line 60: @BeatrixCohere @lusmoura

is there a reason why we consolidate reranking results [here](https://github.com/cohere-ai/cohere-toolkit/blob/main/src/backend/chat/collate.py#L9) instead of localizing it in the tools themselves? I feel it would be reasonable to do the rerank in the tool, as in this example we need a bit more flexibility on how the rerank is implemented.

lusmoura: IMO it's okay to have custom implementations for each tool. But if the tool uses the default chunk and rerank, implementing it with all tool results will lead to less duplicated code

lusmoura: Why do we need a custom implementation in this case?

src/backend/tools/web_scrape.py
```diff
@@ -0,0 +1,28 @@
...
+
+class WebScrapeTool(BaseTool):
+    @classmethod
+    def is_available(cls) -> bool:
+        return True
+
+    def call(self, parameters: dict, **kwargs: Any) -> List[Dict[str, Any]]:
+        url = parameters.get("url")
+
+        page_content = get(url).text
```
lusmoura, line 17: Check if the response was successful before fetching the text. Some pages are restricted, and a GET request may not be sufficient to retrieve the content

```suggestion
        response = get(URL)
        
        if not response.ok:
                return ["text": response.reason] # maybe something else?
         
         page_content = response.text        
```

src/backend/tools/tavily.py
scott-cohere: in this special case we want to rerank -> filter out results from the same url (only want one result from a url) -> return only top 6 results. if we do not do these extra filterings we end up with a ton of references/citations.

src/backend/tools/web_scrape.py
scott-cohere: great callout. thanks

src/backend/tools/tavily.py
lusmoura: Goctha! Maybe we should add a flag to indicate these documents shouldn't be reranked again with all the tool results?

scott-cohere: Revisiting this now. I think its fine that it gets reranked in `custom_chat` since that rerank function there reranks all tool_results and not for the specific tool.

scott-cohere:
closing in favour of https://github.com/cohere-ai/cohere-toolkit/pull/393

PR #289, opened 2024-06-25
frontend: Resolve the NavigationUserMenu hydration error when running without authentication

malexw:
When loading a page while authentication is disabled on the backend we would often see a error message complaining that the rehydrated page differs from the page rendered server-side. This change causes the NavigationUserMenu to always only ever be loaded after the initial render and prevents the error.

**AI Description**

This pull request updates the `src/interfaces/coral_web` directory, specifically the `Layout.tsx`, `authConfig.ts`, and `session.ts` files. 

## Summary

The changes in this PR are related to the react imports and the `useSession` hook. 

## Changes

- In `Layout.tsx`, the React import statement has been updated to include `useState` and `useEffect`.
- The `userMenu` state is now initialized using `useState` and is updated in the `useEffect` hook when the `session` changes.
- The `authConfig.ts` file has some import statements reordered.
- In `session.ts`, the `useSession` hook now uses `useMemo` to memoize the `session` value, which depends on the `authToken`.


tianjing-li (approved):
Beautiful

PR #301, opened 2024-06-26
coral-web: support multihop tool streaming events

wujessica:
**Description:** 
This PR shows the tool events streaming in when using multihop. It also updates some tool names to match what the backend gives us.

Note: the stream response doesn't complete, likely due to my local setup being a bit broken but this demonstrates parsing the tool events streaming in:

https://github.com/cohere-ai/cohere-toolkit/assets/5898755/e659cb09-588d-470c-a840-75272aee25c4

**AI Description**

This PR updates the code to replace references to `TOOL_INTERNET_SEARCH_ID` with `TOOL_WEB_SEARCH_ID`. This includes changes in the following files:

- `src/interfaces/coral_web/src/components/Citations/CitationDocumentHeader.tsx`
- `src/interfaces/coral_web/src/components/FirstTurnSuggestions.tsx`
- `src/interfaces/coral_web/src/components/ToolEvents.tsx`
- `src/interfaces/coral_web/src/constants.ts`
- `src/interfaces/coral_web/src/hooks/chat.ts`
- `src/interfaces/coral_web/src/types/message.ts`

Additionally, there are some other minor changes, such as updating the default value of `isStepsExpanded` in `src/interfaces/coral_web/src/components/MessageRow.tsx` and removing some unused imports.

### DIFF
--- a/src/interfaces/coral_web/src/hooks/chat.ts
+++ b/src/interfaces/coral_web/src/hooks/chat.ts
@@ -14,8 +14,8 @@ import {
   StreamSearchResults,
   StreamStart,
   StreamTextGeneration,
-  StreamToolInput,
-  StreamToolResult,
+  StreamToolCallsChunk,
+  StreamToolCallsGeneration,
   isCohereNetworkError,
   isSessionUnavailableError,
   isStreamError,
@@ -202,7 +202,11 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
     let citations: Citation[] = [];
     let documentsMap: IdToDocument = {};
     let outputFiles: OutputFiles = {};
-    let toolEvents: StreamToolInput[] = [];
+    let toolEvents: StreamToolCallsGeneration[] = [];
+    let currentToolEventIndex = 0;
+
+    // Temporarily store the streaming `parameters` partial JSON string for a tool call
+    let toolCallParamaterStr = '';
 
     try {
       clearComposerFiles();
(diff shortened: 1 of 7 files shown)


src/interfaces/coral_web/src/hooks/chat.ts
```diff
@@ -248,9 +252,59 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
...
+
+                if (data?.tool_call_delta?.name) {
+                  toolCall.name = data.tool_call_delta.name;
+                }
+                if (data?.tool_call_delta?.parameters) {
+                  toolCallParamaterStr += data?.tool_call_delta?.parameters;
+
+                  // Attempt to parse the partial parameter string as valid JSON to show that the parameters
+                  // are streaming in. To make the partial JSON string valid JSON after the object key comes in,
+                  // we naively try to add `"}` to the end.
```
wujessica, line 285: This is a bit janky but unfortunately we don't receive a structured object like we do in Coral i.e. 
```
data: {"event": "tool-calls-chunk", "data": {"tool_call_delta": {"name": null, "index": 0, "parameters": "{\n            \""}, "text": ""}}

data: {"event": "tool-calls-chunk", "data": {"tool_call_delta": {"name": null, "index": 0, "parameters": "query"}, "text": ""}}

data: {"event": "tool-calls-chunk", "data": {"tool_call_delta": {"name": null, "index": 0, "parameters": "\":"}, "text": ""}}

data: {"event": "tool-calls-chunk", "data": {"tool_call_delta": {"name": null, "index": 0, "parameters": " \""}, "text": ""}}

data: {"event": "tool-calls-chunk", "data": {"tool_call_delta": {"name": null, "index": 0, "parameters": "average"}, "text": ""}}
```

vs

```
{
  "result": {
    "eventType": "STREAM_EVENT_TYPE_TOOL_CALLS_CHUNK",
    "isFinished": false,
    "toolCallsChunkStreamEvent": {
      "toolCallDelta": {
        "parameter": {
          "name": "query"
        }
      }
    }
  }
}{
  "result": {
    "eventType": "STREAM_EVENT_TYPE_TOOL_CALLS_CHUNK",
    "isFinished": false,
    "toolCallsChunkStreamEvent": {
      "toolCallDelta": {
        "parameter": {
          "name": "query",
          "value": "\""
        }
      }
    }
  }
}{
  "result": {
    "eventType": "STREAM_EVENT_TYPE_TOOL_CALLS_CHUNK",
    "isFinished": false,
    "toolCallsChunkStreamEvent": {
      "toolCallDelta": {
        "parameter": {
          "name": "query",
          "value": "average"
        }
      }
    }
  }
}
```

misspia-cohere: :')

src/interfaces/coral_web/src/hooks/chat.ts
```diff
@@ -248,9 +252,59 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
...
+                  try {
+                    const partialParams = JSON.parse(toolCallParamaterStr + `"}`);
+                    toolCall.parameters = partialParams;
+                  } catch (e) {
+                    // Ignore parsing error
```
misspia-cohere, line 290: should we report this somewhere to model/BE can fix common issues

misspia-cohere (approved):
🐸

src/interfaces/coral_web/src/hooks/chat.ts
wujessica: i don't think we have any reporting structure as of now 😅 but this should self-heal in the `TOOL_CALLS_GENERATION` event if anything. this is also mostly only in place because of the fact that we're receiving partial JSON strings so the appending of `"}` is guaranteed to work only while the param value is being streamed in and not any other part of the JSON. I have a feeling if there is a problem with the JSON itself it'll probably happen in `TOOL_CALLS_GENERATION` and it'll make the stream return early with an error.

PR #320, opened 2024-06-28
[backend] refactor tools metadata schema

scott-cohere:
1. change artifact schema to be stored as json, would look like this in payload 
```
[
    {
        "user_id": "user-id",
        "organization_id": null,
        "id": "d972bab4-7877-47f1-892b-2c7ee34c32f3",
        "created_at": "2024-06-28T17:41:40.562919",
        "updated_at": "2024-06-28T17:41:40.562919",
        "version": 1,
        "name": "scott's assistant 3",
        "description": "",
        "preamble": "answer everything as a pirate",
        "temperature": 0.0,
        "tools": [
            "web_search"
        ],
        "tools_metadata": [
            {
                "user_id": "user-id",
                "organization_id": null,
                "id": "8234886e-280e-48eb-906b-7dedc7280a1d",
                "tool_name": "ToolName.Google_Drive",
                "artifacts": [
                    {
                        "id": "23123",
                        "name": "readme.md",
                        "type": "file"
                    },
                    {
                        "id": "123123",
                        "name": "readme",
                        "type": "folder"
                    }
                ]
            }
        ],
        "model": "command-r",
        "deployment": "Cohere Platform"
    }
]
```

2. Also added the tools metadata to GET agents

**AI Description**

This PR makes changes to the `agent_tool_metadata` table. 

- The `type` column is dropped.
- The `artifacts` column is dropped and a new column with the same name is added, changing the data type from `ARRAY(TEXT)` to `ARRAY(JSONB)`.
- The unique constraint `_user_agent_tool_name_type_uc` is dropped and a new one, `_user_agent_tool_name_uc`, is created without the `type` column.

The changes are also reflected in the `AgentToolMetadata` model and related test files.

- The `type` attribute is removed from the `AgentToolMetadata` model and related test files.
- The `artifacts` attribute is updated to a list of dictionaries in the `AgentToolMetadata` model and related test files.
- The `Agent` model now includes a `tools_metadata` attribute, which is a list of `AgentToolMetadata`.
- The `AgentToolMetadata` model is imported in the `agent.py` file.
- The `generate:client` script in `package.json` now uses `0.0.0.0` as the IP address instead of `localhost`.
- The `AgentToolMetadata` type in `Agent.ts` now includes an import statement and the `tools_metadata` attribute is added to the `Agent` type.
- The `type` attribute is removed from the `AgentToolMetadata` type in `AgentToolMetadata.ts`, and the `artifacts` attribute is updated to an array of records.
- The `type` attribute is removed from the `CreateAgentToolMetadata` type in `CreateAgentToolMetadata.ts`, and the `artifacts` attribute is updated to an array of records.
- The `type` attribute is made optional in the `UpdateAgentToolMetadata` type in `UpdateAgentToolMetadata.ts`, and the `artifacts` attribute is updated to accept null values.


src/backend/alembic/versions/a48691a80366_.py
```diff
@@ -0,0 +1,74 @@
...
+
+    op.drop_column("agent_tool_metadata", "artifacts")
+    op.add_column(
+        "agent_tool_metadata",
+        sa.Column(
+            "artifacts",
+            postgresql.ARRAY(postgresql.JSONB(astext_type=sa.Text())),
+            nullable=True,
+        ),
+    )
```
scott-cohere, line 37: I had to manually alter this. Changing the column type in the alembic db_model file will fail since there it exists as TEXT[] and you can't marshal it to JSON[] without doing something in the script itself here. 

Should be fine when we deploy since this feature isn't used yet. Lmk otherwise.

@lusmoura @tianjing-li

rodrigue-h (approved):

PR #322, opened 2024-06-28
Add build targets for backend / frontend

rodrigue-h:
Adding build targets for frontend & backend.

### DIFF
--- a/src/backend/Dockerfile
+++ b/src/backend/Dockerfile
@@ -50,5 +50,5 @@ CMD uvicorn backend.main:app --reload --host 0.0.0.0 --port ${PORT}
 
 # production specific target
 FROM base as prod 
-RUN pip install  gunicorn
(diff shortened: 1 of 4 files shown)


src/backend/Dockerfile
```diff
@@ -47,3 +47,8 @@ COPY .en[v] .env
 EXPOSE ${PORT}
 
 CMD uvicorn backend.main:app --reload --host 0.0.0.0 --port ${PORT}
+
+# production specific target
+FROM base as prod 
+RUN pip install  gunicorn
```
marwanad, line 53: id pin the gunicorn version

marwanad (approved):
nothing blocking, seems sane

src/backend/Dockerfile
rodrigue-h: done!

PR #327, opened 2024-07-01
frontend(chore): update ci dependencies

tomtobac:
## Description

- fix: run ci build on every pr against `main`
- chore: force to use `npm` as package manager
- chore: bump node version to up 20.12.2
- refactor: simplify frontend docker file
- chore: remove babel, env-cmd  dependency since are not longer needed

Closes OS-2208


lusmoura (approved):

knajjars (approved):
can we add `pnpm-lock.yaml` and `yarn.lock` to the .gitignore?

PR #336, opened 2024-07-02
frontend: generate title after conversation starts

tomtobac:
### Description

Generate the conversation title after the conversation starts and after 5 iterations.


src/interfaces/coral_web/src/hooks/chat.ts
```diff
@@ -442,25 +468,24 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
                 toolEvents,
               });
 
+              if (shouldUpdateConversationTitle(newMessages)) {
+                handleUpdateConversationTitle(conversationId);
```
knajjars, line 472: do we want to `await` this?

knajjars: probably no because of the timeout

knajjars (approved):

PR #341, opened 2024-07-02
Conversation: add rerank search

lusmoura:

**AI Description**

### DIFF
--- a/src/backend/routers/conversation.py
+++ b/src/backend/routers/conversation.py
@@ -26,10 +26,12 @@ from backend.services.chat import generate_chat_response, get_deployment_config
 from backend.services.conversation import (
     DEFAULT_TITLE,
     GENERATE_TITLE_PROMPT,
+    SEARCH_RELEVANCE_THRESHOLD,
     extract_details_from_conversation,
 )
 from backend.services.file.service import FileService
 from backend.tools.files import get_file_content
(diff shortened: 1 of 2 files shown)


src/backend/routers/conversation.py
```diff
@@ -167,6 +169,80 @@ async def delete_conversation(
...
+        filtered_conversations = []
+
+        for rerank_document, conversation in zip(rerank_documents, conversations):
+            if query.lower() in rerank_document.lower():
+                filtered_conversations.append(conversation)
+
+        return filtered_conversations
+
+    # Rerank documents
+    res = model_deployment.invoke_rerank(
```
tianjing-li, line 225: We should do some validation on the `res` here since not all model deployments have a proper `invoke_rerank` method. Good example is `BedrockDeployment` which just returns None

lusmoura: I think all the model deployments have a flag `rerank_enabled` which I'm using to check that in the condition above, but I'll add some extra validation here just in case

tianjing-li (approved):
lgtm

src/backend/routers/conversation.py
tianjing-li: @lusmoura ah yep good point, thanks for adding validation

PR #348, opened 2024-07-03
Metrics: fix error when data is None and improve logger

lusmoura:

**AI Description**

The pull request makes changes to the `metrics.py` file in the backend services of a Python application. The modifications aim to improve error handling, logging, and asynchronous operations related to metrics collection and reporting.

Here's a summary of the changes:
- Added imports for `get_header_user_id` and `logging` modules.
- Introduced a logger instance with the error-level severity.
- Modified the `dispatch` method to use `await` with `run_loop` to ensure asynchronous execution.
- Updated the `get_event_data` method to use a `try-except` block for obtaining the `user_id` using `get_header_user_id`. It also logs a warning if getting the user ID fails.
- Refactored multiple methods (`get_method`, `get_endpoint_name`, `get_status_code`, `get_success`, `get_user_id`, `get_user`, `get_object_ids`, and `get_agent') to log warnings using the logger instance instead of `logging`.
- Added a check in the `report_metrics` method to ensure that metrics data is not empty before proceeding with reporting.
- Modified the `report_metrics` method to use the logger instance for error logging instead of `logging`.
- Updated the `run_loop` method to be an asynchronous function, including error handling using `task_exception_handler` to catch and log exceptions.
- Modified the `collect_metrics_chat_stream` and `collect_metrics_rerank` decorators to use the `await` keyword with `run_loop` to ensure asynchronous execution.

### DIFF
--- a/src/backend/services/metrics.py
+++ b/src/backend/services/metrics.py
@@ -175,8 +175,7 @@ class MetricsMiddleware(BaseHTTPMiddleware):
 
 async def report_metrics(data: MetricsData) -> None:
     if not data:
-        logger.error("No metrics data to report")
-        return
+        raise ValueError("No metrics data to report")
 
     data = attach_secret(data)
     signal = MetricsSignal(signal=data)
@@ -184,6 +183,7 @@ async def report_metrics(data: MetricsData) -> None:
     if not REPORT_SECRET:
         logger.error("No report secret set")
         return
+    
     if not REPORT_ENDPOINT:
         logger.error("No report endpoint set")
         return
--- a/src/backend/services/metrics.py
+++ b/src/backend/services/metrics.py
@@ -23,12 +23,15 @@ from backend.schemas.metrics import (
     MetricsSignal,
     MetricsUser,
 )
+from backend.services.auth.utils import get_header_user_id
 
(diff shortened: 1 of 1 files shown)


mtanzim (approved):
please consider logging the sources of empty data so we can fix incrementally

src/backend/services/metrics.py
```diff
@@ -168,6 +168,9 @@ def get_agent(self, request: Request) -> Union[MetricsAgent, None]:
 
 
 async def report_metrics(data: MetricsData) -> None:
+    if not data:
```
tanzimCohere, line 171:
```python
if not data:
    throw Exception("empty data")
```

src/backend/services/metrics.py
```diff
@@ -168,14 +174,18 @@ def get_agent(self, request: Request) -> Union[MetricsAgent, None]:
...
+        logger.error("No metrics data to report")
```
mtanzim, line 178: should we *throw* an exception here so we can trace back to which parts are sending empty data?

mtanzim: as I understand, this code will log the error when it is thrown:
https://github.com/cohere-ai/cohere-toolkit/blob/d75f8083ed18fa7dac0b20d58f950aeaf053b44d/src/backend/services/metrics.py#L224-L225

PR #354, opened 2024-07-04
Files: support more file types and improve parsing

lusmoura:
- Add support to more file types and raise error on invalid type
- Sanitize file names so the model always generates a valid name
- Remove old File service and improve file handling -- don't save file locally anymore

**AI Description**

This PR introduces changes to the file handling system. The FileService class has been removed, and the file handling logic has been refactored. 

Here are the key changes:
- The FileService class and its methods `upload_file`, `delete_file`, and `generate_new_filepath' have been removed.
- The `get_file_content` function has been updated to support various file formats (PDF, text, markdown, CSV, JSON, and DOCX). It now uses the new `get_file_extension` helper function to determine the file type.
- A new function, `read_excel`, has been added to handle Excel files.
- The `validate_file_size` function has been introduced to validate the file size before uploading. It checks if the file size exceeds the maximum allowed limit (20MB) and if the total file size per user exceeds 1GB.
- The `get_files_by_user_id` function has been added to list all files by a user's ID.
- The `upload_file` function now uses the `get_file_content` and `validate_file_size` functions. It also replaces null characters (`\x00`) in the file content and encodes the filename to ASCII.
- The `delete_file` function has been updated to only delete the File DB object, removing the local file deletion logic.
- The `get_files_by_file_names` function now decodes the file names to UTF-8.
- The `test_list_files_by_user_id` and `test_list_files_by_user_id_empty` test cases have been added to test the `get_files_by_user_id` function.
- The `openpyxl` package has been added to the `pyproject.toml` file.


pyproject.toml
```diff
@@ -44,6 +44,8 @@ transformers = "^4.41.2"
 google_auth_oauthlib ="^1.2.0"
 google-auth-httplib2="^0.2.0"
 google-api-python-client="^2.133.0"
+pypdf2 = "^3.0.1"
```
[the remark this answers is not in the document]
lusmoura: The reported issues are unrelated to this! But I'll fix the declaration

src/backend/services/file/service.py
```diff
@@ -1,95 +1,56 @@
...
+async def get_file_content(file: FastAPIUploadFile) -> str:
+    file_contents = await file.read()
+    file_extention = get_file_extention(file.filename)
 
-        Args:
-            file_path (Path): File path.
+    if file_extention == PDF_EXTENTION:
+        return read_pdf(file_contents)
+    elif file_extention in [
+        TEXT_EXTENTION,
```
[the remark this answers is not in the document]
lusmoura: Yep, at least as far as I'm aware

src/backend/routers/conversation.py
```diff
@@ -220,37 +219,29 @@ async def upload_file(
...
         upload_file = FileModel(
             user_id=conversation.user_id,
             conversation_id=conversation.id,
-            file_name=file_path.name,
-            file_path=str(file_path),
-            file_size=file_path.stat().st_size,
-            file_content=content,
+            file_name=filename,
+            file_path=filename,
+            file_size=file.size,
```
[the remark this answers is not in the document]
lusmoura: Sure! I'll limit the filesize to 20Mb and the total file size to 1Gb

src/backend/services/conversation.py
```diff
@@ -11,6 +16,8 @@
 
 # TITLE
 """
+MAX_FILE_SIZE = 20_000_000  # 20MB
+MAX_TOTAL_FILE_SIZE = 1_000_000_000  # 1GB
```
tianjing-li, line 20: TIL you could do this!

src/backend/services/conversation.py
```diff
@@ -52,3 +59,31 @@ def extract_details_from_conversation(
 
     chatlog = "\n".join(turns)
     return chatlog
+
+
+def validate_file_size(
```
tianjing-li, line 64: Perhaps this method and the above constant should go in the file/service?

src/backend/services/file/service.py
```diff
@@ -1,95 +1,56 @@
-import os
-import re
```
tianjing-li, line 2: I realize I implemented this in the past, but let's remove the `file` folder, rename the `service.py` to `file.py` and move it in the `services` folder

src/backend/services/file/service.py
```diff
@@ -1,95 +1,56 @@
...
+import io
 from typing import Any
 
+import pandas as pd
+from fastapi import UploadFile as FastAPIUploadFile
+from pypdf import PdfReader
 
-class FileService:
-    DEFAULT_DATA_FOLDER = "src/backend/data"
+PDF_EXTENTION = "pdf"
```
tianjing-li, line 8: Nit: these should be EXTENSION

src/backend/services/file/service.py
```diff
@@ -1,95 +1,56 @@
...
-    def generate_new_filepath(self, file_path: Path) -> Path:
-        """
-        Generate a new file path. The path is generated by concatenating the base name with a counter index.
+async def get_file_content(file: FastAPIUploadFile) -> str:
+    file_contents = await file.read()
+    file_extention = get_file_extention(file.filename)
 
-        Args:
-            file_path (Path): File path.
+    if file_extention == PDF_EXTENTION:
```
tianjing-li, line 26: Also nit to just change the wording from `extention` to `extension` where applicable

tianjing-li (changes requested):
Couple changes requested

tianjing-li (approved):
LGTM - thanks for changes

PR #366, opened 2024-07-05
backend: Add global app-level error handler

tianjing-li:
This should catch in endpoints any exceptions that are uncaught, and return a better error message response + log.


Makefile
```diff
@@ -3,7 +3,7 @@ dev:
 down:
 	@docker compose down
 run-tests:
-	docker compose run --build backend poetry run pytest src/backend/tests/$(file)
+	docker compose run --build backend poetry run pytest src/backend/tests/routers/test_main.py
```
lusmoura, line 6: nit: remove this change

src/backend/main.py
```diff
@@ -82,6 +83,23 @@ def create_app():
 app = create_app()
 
 
+@app.exception_handler(Exception)
+async def validation_exception_handler(request: Request, exc: Exception):
+    logger.info(
+        f"Error occurred: {exc!r} during request: {request.method}, {request.url}"
```
lusmoura, line 89: TIL!

lusmoura (approved):

Makefile
tianjing-li: done! ty

PR #371, opened 2024-07-08
feat(toolkit): Dedupe links when URL is cited multiple times in the references

knajjars:
https://github.com/cohere-ai/cohere-toolkit/assets/19353218/0e4c0e33-ea27-4cee-8cf5-56fcb139bd37

**AI Description**

This pull request makes changes to the `Citation.tsx` component in the `src/interfaces/coral_web/src/components/Citations` directory. 

## Summary
The `Citation` component renders a list of citations with associated documents. The changes in this PR modify the logic for displaying citations and documents based on selection state and visibility settings. 

## Changes
- Updated the condition to return `null` when there are no relevant documents or when a citation is not selected.
- Introduced a new variable, `uniqueDocumentsUrls`, to store unique document URLs.
- Modified the condition for rendering the number of references and the associated icon button. Now, it considers the length of `uniqueDocumentsUrls` instead of `uniqueDocuments`.
- Refactored the mapping logic for rendering individual documents:
  - When a citation is selected, it maps over `uniqueDocuments` and renders only the selected documents.
  - When no citation is selected, it maps over `uniqueDocumentsUrls` and renders documents based on visibility settings.

### DIFF
--- a/src/interfaces/coral_web/src/components/Citations/Citation.tsx
+++ b/src/interfaces/coral_web/src/components/Citations/Citation.tsx
@@ -69,7 +69,7 @@ export const Citation = React.forwardRef<HTMLDivElement, Props>(function Citatio
     citationRef: containerRef,
   });
 
-  if (!startEndKeyToDocs || documents.length === 0) {
+  if (!startEndKeyToDocs || documents.length === 0 || (!isSelected && !!selectedCitation)) {
     return null;
   }
 
@@ -80,6 +80,7 @@ export const Citation = React.forwardRef<HTMLDivElement, Props>(function Citatio
     uniqBy(flatten(Object.values(startEndKeyToDocs)), 'document_id'),
     'document_id'
   );
+  const uniqueDocumentsUrls = uniqBy(uniqueDocuments, 'url');
 
   const handleMouseEnter = () => {
     hoverCitation(generationId);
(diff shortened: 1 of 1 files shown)


src/interfaces/coral_web/src/components/Citations/Citation.tsx
```diff
@@ -146,9 +147,9 @@ export const Citation = React.forwardRef<HTMLDivElement, Props>(function Citatio
 
         <div className={cn('mb-4 flex items-center justify-between', { hidden: isSelected })}>
           <Text as="span" styleAs="caption" className="text-volcanic-800">
-            {uniqueDocuments.length} {pluralize('reference', uniqueDocuments.length)}
+            {uniqueDocuments.length} {pluralize('reference', uniqueDocuments.length)} from
```
abimacarmes, line 150: Would it be worth making this slightly more clear that it's XX references from XX urls/documents?

src/interfaces/coral_web/src/components/Citations/Citation.tsx
abimacarmes, line 1: It's still showing the duplicated links when selected (0:06 of the video) - it's different sections of the same text but would we want to have two different preview chunks with a single title?

abimacarmes (approved):

src/interfaces/coral_web/src/components/Citations/Citation.tsx
knajjars: I think we want to keep it generic because this could be a combination of docs, urls, tools, etc.

PR #376, opened 2024-07-08
metrics: add event types

mtanzim:

**AI Description**

This PR introduces changes to the metrics system and authentication configuration. 

**Files changed:**
- `helper_scripts/metrics_helper.py`
- `src/backend/config/auth.py`
- `src/backend/schemas/metrics.py`
- `src/backend/services/metrics.py`

**Changes:**
- Added `MetricsMessageType` enum in `metrics.py` to define message types for various events.
- Modified `MetricsDataBase` in `metrics.py` to include `message_type` field and removed `method`, `endpoint_name`, and `success` fields.
- Updated `get_event_data` function in `metrics.py` to determine `message_type` based on `method`, `endpoint_name`, and `is_success`.
- Added `ROUTE_MAPPING` dictionary in `metrics.py` to map endpoint paths to `MetricsMessageType`.
- Introduced `event_name_of` function in `metrics.py` to determine the event name based on `method`, `endpoint_name`, and `is_success`.
- Modified `cleanup` function in `metrics_helper.py` to include an `if agent_id` check before deleting agent-related data.
- Changed `ENABLED_AUTH_STRATEGIES` in `auth.py` to an empty list and added a comment to fix it before merging.


src/backend/schemas/metrics.py
```diff
@@ -1,17 +1,38 @@
...
+    UNKNOWN_SIGNAL = "unknown"
+
+
 class MetricsDataBase(BaseModel):
     id: str
     user_id: str
     trace_id: str
-    method: str
-    endpoint_name: str
-    success: bool
```
scott-cohere, line 14: do the new message type encapsulate the success status? for example, i see we have success and fail types but for something like `assistant_created` there is no indication that it succeeded or failed, or is that not important for the metrics use case?

scott-cohere (approved):
small comment

src/backend/schemas/metrics.py
mtanzim: The main reason we sent success before was for the mapping to event name. Without that, the success isn't needed. The goal long term would be to remove the mapping altogether and simply work with event types for the events we care about.

PR #383, opened 2024-07-09
User: require existing user 

lusmoura:
When different users tried to create a conversation with the same ID, DB operations caused a 500. Since the `conversations` table had only the `ID` as PK, there was a duplicate key error. Now, the PK in that table is a combination of `ID` and `user_id`. All tables that had some reference to `conversations` were updated, as well as the unit tests.

Another big change is that every request requiring a `User-Id` has a new constraint to validate the user exists in the DB. Since in `main` there's no auth, FE uses the default user with ID `user_id`, which is now added to the DB when there's a migration.

**AI Description**

This PR makes changes to multiple files, including:

- **.env-template**: Added configuration for Compass and Google Drive.
- **.github/workflows/docker_push_frontend.yml**: Modified the `docker_push_frontend.yml` file.
- **.pre-commit-config.yaml**: Updated the `rev` values for `black` and `ruff-pre-commit'.
- **Makefile**: Changed the command for attaching to the backend container.
- **docker-compose.yml**: Removed the `POSTGRES_DB` variable and added `NEXT_PUBLIC_GOOGLE_DRIVE_CLIENT_ID` and `NEXT_PUBLIC_GOOGLE_DRIVE_DEVELOPER_KEY` to the `environment` section.
- **docs/custom_tool_guides/tool_guide.md**: Modified the `call` function to be asynchronous and updated its return type.
- **helper_scripts/metrics_helper.py**: Added `version', `description', `preamble', and `temperature' to the `create_agent` function. Also, added a check before deleting the agent ID and included a TODO comment for creating tests.
- **poetry.lock**: Updated the version of multiple packages, including `boto3', `botocore', `fsspec', `faker', `openai', `openpyxl', and `tenacity'. Additionally, added the `et-xmlfile` package.
- **pyproject.toml**: Added `openpyxl` and the Compass dependencies section.
- **src/backend/Dockerfile**: Added a new line to the Dockerfile.
- **src/backend/alembic/versions/1acff4138b79_.py**: Added a new migration file to update the `user_id` column in multiple tables.
- **src/backend/alembic/versions/ed17f144f4bf_.py**: Added a new migration file to populate the `users` table with the default user.
- **src/backend/chat/collate.py**: Modified the `rerank_and_chunk` function to be asynchronous.
- **src/backend/chat/custom/custom.py**: Imported `send_log_message` and modified the `chat` and `call_chat` functions to be asynchronous. Also, added `send_log_message` calls and removed the `print` statement.
- **src/backend/compass_sdk/__init__.py**: Added a new module for the Compass SDK, including classes for logging, validated models, Compass-specific data structures, and configuration.
- **src/backend/compass_sdk/compass.py**: Added a new module for the Compass SDK, including a client class for interacting with the Compass API.
- **src/backend/compass_sdk/constants.py**: Added a new module for the Compass SDK, defining default values and constants.
- **src/backend/compass_sdk/parser.py**: Added a new module for the Compass SDK, including a client class for interacting with the Compass Parser API.
- **src/backend/compass_sdk/utils.py**: Added a new module for the Compass SDK, providing utility functions for working with files and file systems.
- **src/backend/config/auth.py**: Added the `SKIP_AUTH` environment variable and enabled the `OpenIDConnect` authentication strategy.
- **src/backend/config/tools.py**: Set the `is_visible` attribute of the `GoogleDrive` tool to `True'.
- **src/backend/crud/file.py**: Added a new function to list files by user ID.
- **src/backend/database_models/agent.py**: Updated the `user_id` column to be a foreign key referencing the `users` table.
- **src/backend/database_models/agent_tool_metadata.py**: Updated the `user_id` column to be a foreign key referencing the `users` table.
- **src/backend/database_models/citation.py**: Updated the `user_id` column to be a foreign key referencing the `users` table.
- **src/backend/database_models/conversation.py**: Added a unique constraint and a primary key constraint to the `id` and `user_id` columns. Also, updated the `user_id` column to be a foreign key referencing the `users` table.
- **src/backend/database_models/document.py**: Added a foreign key constraint to the `conversation_id` and `user_id` columns.
- **src/backend/database_models/file.py**: Added a foreign key constraint to the `conversation_id` and `user_id` columns.


src/backend/database_models/conversation.py
```diff
@@ -11,8 +12,8 @@
 class Conversation(Base):
     __tablename__ = "conversations"
 
-    # TODO: Swap to foreign key once User management implemented
-    user_id: Mapped[str] = mapped_column(String)
+    id: Mapped[str] = mapped_column(String, default=lambda: str(uuid4()))
```
lusmoura, line 15: We need to override the ID to be able to use it as a composite primary key

src/backend/alembic/versions/3a2c90817791_.py
```diff
@@ -0,0 +1,155 @@
...
+        "documents_conversation_id_fkey", "documents", type_="foreignkey"
+    )
+    op.drop_constraint("files_user_id_fkey", "files", type_="foreignkey")
+    op.drop_constraint("messages_conversation_id_fkey", "messages", type_="foreignkey")
+    op.drop_constraint("messages_user_id_fkey", "messages", type_="foreignkey")
+    op.drop_constraint("snapshots_user_id_fkey", "snapshots", type_="foreignkey")
+    op.drop_constraint(
+        "snapshots_conversation_id_fkey", "snapshots", type_="foreignkey"
+    )
+    op.drop_constraint("conversations_pkey", "conversations", type_="primary")
```
lusmoura, line 35: Had to manually update the migration because Alembic doesn't support updating primary keys [[ref](https://stackoverflow.com/questions/13756567/adding-primary-key-to-existing-mysql-table-in-alembic)]

tianjing-li (changes requested):
I think the default user changes are great, but have some concerns about the new composite key of id + user_id on Conversations

tianjing-li (approved):
Thanks for explaining, let's go with this solution.

src/backend/config/deployments.py
```diff
@@ -44,27 +44,27 @@ class ModelDeploymentName(StrEnum):
...
-        env_vars=AZURE_ENV_VARS,
-    ),
-    ModelDeploymentName.Bedrock: Deployment(
-        name=ModelDeploymentName.Bedrock,
-        deployment_class=BedrockDeployment,
-        models=BedrockDeployment.list_models(),
-        is_available=BedrockDeployment.is_available(),
-        env_vars=BEDROCK_ENV_VARS,
-    ),
+    # ModelDeploymentName.SageMaker: Deployment(
```
tianjing-li, line 47: Is it intentional to keep these commented out?

tianjing-li (approved):
Looks good, just one comment

PR #385, opened 2024-07-09
slack-bot: Add commands to set and view deployment

scottmx81:
**Description:**

This PR adds two new commands to the slack bot: `/set-deployment` and `/view-deployment`. The deployment can be set per channel, and the value is stored in the PostgreSQL db. Every call to the toolkit API from the Slack bot is modified to send a `Deployment-Name:` header, with the value that the user passed in the `/set-deployment` command.

The Slack bot uses an auto-generated API client, and the current `openapi.json` file before this PR, does not document the `Deployment-Name` header, meaning there was no way to pass it through the auto-generated client cleanly. The only option would have been to set it globally in the `OpenAPI` config object that is passed to the `ToolkitClient` constructor, but changing it in the config option is not ideal here, since each channel can have its own deployment value set.

In order to be able to pass the `Deployment-Name` header in the call to the toolkit client's `chatV1ChatPost` function, I had to make sure the `Deployment-Name` header was included in the generated `openapi.json` file. This PR changes function signature of the `routers.chat.chat` function to include a `deployment_name` arg. It's purpose was to get the header in the `openapi.json` file, even though it appears to be unused.

When I regenerated the toolkit client from the updated spec, there were other things that changed also. This required several changes to the Slack bot code that do not appear related to the `/set-deployment` and `/view-deployment` commands, but were necessary to get regenerated toolkit client fields to work and not have any typing errors or warnings.

Additionally, while working on this feature, I realized that a dependency was updated by a dependabot PR, which actually broke the PDF parsing features in the Slack bot, without getting caught by tests. I have downgraded the `pdfjs-dist` dependency to the last 3.x release, because the 4.x releases are not compatible without changes that are outside the scope of this PR.

If no deployment has been set, or if the user runs the `/set-deployment` command without any arguments, it will default to using `Cohere Platform` as the deployment.

Issues that I am aware of that have not been addressed yet, due to time constraints and which should be considered in the future:

* There are no restrictions on who can run the `/set-deployment` and `/set-model` commands. We may want to lock this down to only workspace admins or channel owners.
* When an invalid value is sent in the `Deployment-Name` header, it gets a 404 from the toolkit API. The slack bot shows a warning reaction to let the user something went wrong, but it would probably be better if a more descriptive error were passed back to the user or at least shown in the logs to say that the deployment is invalid.
* It does not currently validate the value passed to `/set-deployment` by the user. It would be nice if it validated the value at the time it is set by the user.

**AI Description**

This PR adds the ability to set and view the deployment used for a specific channel. It also includes changes to the underlying code to support this new functionality.

## Summary
- Adds the ability to set and view the deployment used for a specific channel.
- Updates the underlying code to support the new functionality.

## Changes
- Adds the /set-deployment and /view-deployment commands.
- Updates the `Chat` endpoint to include the `deployment_name` parameter.
- Adds the `deployment` field to the `ChannelSettings` model.
- Updates the `ChatMessage` and `ChatRole` enums to include the `SYSTEM` role.
- Updates the `Agent` and `Snapshot` models to include the `deployment` field.
- Updates the `CreateAgent` and `UpdateAgent` models to include the `deployment` field.
- Updates the `DefaultService` class to include the `setDeployment` and `viewDeployment` methods.


src/backend/routers/chat.py
```diff
@@ -97,6 +97,7 @@ async def chat(
     chat_request: CohereChatRequest,
     request: Request,
     agent_id: str | None = None,
+    deployment_name: str | None = Header(default=None),
```
tianjing-li, line 100: Can we add a # TODO comment to have request validator use this param?

scottmx81: @tianjing-li I have added a comment as requested.

tianjing-li (approved):
lgtm

PR #393, opened 2024-07-10
backend: web scraping in tools + fix metrics data use in tools

scott-cohere:
- Adding scraping to web search
- Adding web scrape tool
- Fix kwargs passing that has been causing chat and rerank calls to fail

image

**AI Description**

This PR makes changes to the following:
- Environment variables
- Docker configuration
- Pre-commit hooks
- Makefile commands
- Dependency versions
- Function signatures
- Function implementations
- Model deployments
- Tool definitions
- Database queries

## Environment Variables
The following environment variables have been added:
- COHERE_COMPASS_USERNAME
- COHERE_COMPASS_PASSWORD
- COHERE_COMPASS_API_URL
- <co: 0>COHERE_COMPASS_PARSER_URL

### DIFF
--- a/src/backend/services/metrics.py
+++ b/src/backend/services/metrics.py
@@ -372,9 +372,9 @@ def initialize_sdk_metrics_data(
         MetricsData(
             id=str(uuid.uuid4()),
             message_type=message_type,
-            trace_id=kwargs.get("trace_id", "unknown"),
-            user_id=kwargs.get("user_id", "unknown"),
-            assistant_id=kwargs.get("agent_id", "unknown"),
+            trace_id=kwargs.get("trace_id", None),
+            user_id=kwargs.get("user_id", None),
+            assistant_id=kwargs.get("agent_id", None),
             model=chat_request.model if chat_request else None,
         ),
         kwargs,
--- a/src/backend/services/metrics.py
+++ b/src/backend/services/metrics.py
@@ -372,9 +372,9 @@ def initialize_sdk_metrics_data(
         MetricsData(
             id=str(uuid.uuid4()),
             message_type=message_type,
-            trace_id=kwargs.get("trace_id", None),
-            user_id=kwargs.get("user_id", None),
-            assistant_id=kwargs.get("agent_id", None),
+            trace_id=kwargs.get("trace_id", "unknown"),
+            user_id=kwargs.get("user_id", "unknown"),
+            assistant_id=kwargs.get("agent_id", "unknown"),
             model=chat_request.model if chat_request else None,
         ),
         kwargs,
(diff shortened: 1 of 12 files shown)


src/backend/tools/web_scrape.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @classmethod
+    def is_available(cls) -> bool:
+        return True
+
+    async def call(self, parameters: dict, **kwargs: Any) -> List[Dict[str, Any]]:
+        url = parameters.get("url")
+
+        response = get(url)
+        if not response.ok:
```
lusmoura, line 20: Wondering if we should return an exception here or a dict containing the error. The python interpreter, for example, returns the std_err when the code fails to execute, so the model knows the issue and tries to correct it. I'm not sure if this could lead to infinite loops here, but it might be worth trying

src/backend/services/metrics.py
```diff
@@ -372,9 +372,9 @@ def initialize_sdk_metrics_data(
         MetricsData(
             id=str(uuid.uuid4()),
             message_type=message_type,
-            trace_id=kwargs.pop("trace_id", None),
-            user_id=kwargs.pop("user_id", None),
-            assistant_id=kwargs.pop("agent_id", None),
+            trace_id=kwargs.get("trace_id", None),
```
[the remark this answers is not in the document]
scott-cohere: i like that, probably would be easier to debug if its "unknown" since we can ref back to here.

src/backend/tools/web_scrape.py
scott-cohere: hmm thats a great point. lets return dict with error in `text` and see how that works out.

src/backend/tools/web_scrape.py
```diff
@@ -18,7 +18,14 @@ async def call(self, parameters: dict, **kwargs: Any) -> List[Dict[str, Any]]:
 
         response = get(url)
         if not response.ok:
-            return Exception(f"Cannot open and scrape URL {url}")
+            return [
+                (
+                    {
+                        "text": f"Cannot open and scrape URL {url}",
```
lusmoura, line 24: last nit: can you add the reason or the error message to the text as well?

scott-cohere: oop yes!

src/backend/services/metrics.py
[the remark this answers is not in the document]
tanzimCohere: unrelated, but why do we pop the kwargs? would that not mutate for future uses of it leading to unexpected behaviour?

re: "unknown", would we still have a way of catching schema errors when sending through metrics? would it simply being a string cause silent errors?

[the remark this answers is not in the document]
scott-cohere: lets talk about this as a follow up. i will change it back to None and merge this PR. since this change isnt really related to the PR but rather just a piggyback

PR #394, opened 2024-07-11
infra(toolkit): fix deployment target for UI on prod and dev

knajjars:
**AI Description**

This pull request makes changes to the `docker-compose.yml` file and adds a new build target feature to the backend.

## docker-compose.yml
- Adds quotes around the port values.
- Moves the `volumes` property under the `db` service to be consistent with other services.
- Updates the `build` property of the `frontend` service to include a `target` property, which is set to `${BUILD_TARGET:-prod}`.

## src/backend/cli/main.py
- Adds a new `BuildTarget` class with `DEV` and `PROD` as possible values.
- Introduces a `build_target_prompt` function to prompt the user to select a build target.
- Integrates the build target selection into the `start` function, setting the `BUILD_TARGET` secret.

## src/interfaces/coral_web/Dockerfile
- Adds a `dev` stage to the Dockerfile, using `yarn dev` or `npm run dev` based on the presence of `yarn.lock` or `package-lock.json`.
- Updates the `prod` stage to use `yarn start` or `npm run start` if `yarn.lock` or `package-lock.json` are present, and falls back to `pnpm start` or `npm run dev` if they are not present.

### DIFF
--- a/src/interfaces/coral_web/Dockerfile
+++ b/src/interfaces/coral_web/Dockerfile
@@ -27,6 +27,7 @@ COPY .en[v] .env
 # Note: Don't expose ports here, Compose will handle that for us
 
 # Start Next.js in development mode based on the preferred package manager
+FROM base as dev
 CMD \
   if [ -f yarn.lock ]; then yarn dev; \
   elif [ -f package-lock.json ]; then npm run dev; \
@@ -36,5 +37,10 @@ CMD \
 
 
 # Production specifc tareget
-# FROM base AS prod
-# CMD npm run start
+FROM base AS prod
+CMD \
+  if [ -f yarn.lock ]; then yarn start; \
+  elif [ -f package-lock.json ]; then npm run start; \
+  elif [ -f pnpm-lock.yaml ]; then pnpm start; \
+  else npm run dev; \
+  fi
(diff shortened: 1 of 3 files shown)


src/interfaces/coral_web/Dockerfile
```diff
@@ -36,5 +37,10 @@ CMD \
...
 
 # Production specifc tareget
-# FROM base AS prod
-# CMD npm run start
\ No newline at end of file
+FROM base AS prod
+CMD \
+  if [ -f yarn.lock ]; then yarn start; \
+  elif [ -f package-lock.json ]; then npm run start; \
+  elif [ -f pnpm-lock.yaml ]; then pnpm start; \
```
tomtobac, line 44: Can we drop the support for all the other lockers?

knajjars: yes I want to, but will do it in a separate PR, because we have to fix the `readme.md` as well and other files

knajjars: https://linear.app/cohereai/issue/OS-2282/[fe]-remove-yarn-and-pnpm-references-in-toolkit

miguelvr (approved):

PR #396, opened 2024-07-11
Agents: remove user_id and org_id + refactor

lusmoura:
1/n - Will do the same refactor for all the endpoints/schemas/types so that:

1. We don't return `user_id` and `organization_id` when not strictly needed
2. We rollback transactions if they fail
3. We have more readable code and better functions

**AI Description**

### DIFF
--- a/src/backend/crud/agent.py
+++ b/src/backend/crud/agent.py
@@ -1,7 +1,7 @@
 from sqlalchemy.orm import Session
 
 from backend.database_models.agent import Agent
-from backend.schemas.agent import UpdateAgent
+from backend.schemas.agent import UpdateAgentRequest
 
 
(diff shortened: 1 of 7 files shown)


src/backend/crud/agent.py
```diff
@@ -17,9 +17,14 @@ def create_agent(db: Session, agent: Agent) -> Agent:
     Returns:
       Agent: Created agent.
     """
-    db.add(agent)
-    db.commit()
-    db.refresh(agent)
+    try:
```
tianjing-li, line 20: Suggestion: how about making a decorator that wraps every crud transaction in a try:except?

lusmoura: 🤯

lusmoura: I always forget decorators exist hahah

PR #407, opened 2024-07-12
frontend: Add back the Register link if basic auth is enabled

malexw:
Restore the link to the Register page if the backend has been configured for login with a username and password.

Also fixes an issue introduced with the change to the new API client that broke Authorization headers.

**AI Description**

This pull request makes changes to the `src/interfaces/coral_web/src/cohere-client/client.ts` and `src/interfaces/coral_web/src/pages/login.tsx` files.

## Summary
The changes in this PR mainly involve updating the `HEADERS` property in the `CohereClient` class and modifying the `LoginPage` component in the `login.tsx` file.

## Changes
- Modified the `HEADERS` property in the `CohereClient` class to be an async function that returns the headers.
- Added the `AuthLink` component import from `'@/components/AuthLink''.
- Introduced `welcomePageProps` to hold the `title` and `navigationAction` props for the `WelcomePage` component.
- Replaced the `WelcomePage` component invocation with its JSXS spread syntax, passing in the `welcomePageProps`.
- Wrapped the form and the new user text with the `AuthLink` component in a fragment `<>`.

### DIFF
--- a/src/interfaces/coral_web/src/pages/login.tsx
+++ b/src/interfaces/coral_web/src/pages/login.tsx
@@ -5,6 +5,7 @@ import { useMemo, useState } from 'react';
 import { SubmitHandler, useForm } from 'react-hook-form';
 
 import { CohereClient, CohereUnauthorizedError, ListAuthStrategy } from '@/cohere-client';
+import { AuthLink } from '@/components/AuthLink';
 import { Button, Input, Text } from '@/components/Shared';
 import { OidcSSOButton } from '@/components/Welcome/OidcSSOButton';
 import { WelcomePage } from '@/components/WelcomePage';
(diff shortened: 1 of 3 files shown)


src/interfaces/coral_web/src/pages/login.tsx
```diff
@@ -116,54 +122,68 @@ const LoginPage: NextPage<Props> = () => {
...
-
-            <Button
-              disabled={loginStatus === 'pending' || !formState.isValid}
-              label={loginStatus === 'pending' ? 'Logging in...' : 'Log in'}
-              type="submit"
-              className="mt-10 w-full self-center md:w-fit"
-              splitIcon="arrow-right"
-            />
-          </form>
+          <>
```
tomtobac, line 125: Is there any reason why we added this Fragment here? 🤔

malexw: With the conditional render opened by `{hasBasicAuth && (`, you're required to have one parent element. Without it, the `<form>` and `<Text>` sibling elements would cause problems with ReactDOM

tianjing-li (approved):
lgtm

PR #435, opened 2024-07-16
backend: Change the JWT expiry time to 3 months

malexw:
Logging in every 12 hours appears to be too short of a default, so we're bumping it up to 3 months.

**AI Description**


tianjing-li (approved):
lgtm

PR #437, opened 2024-07-16
metrics: send the correct time from event

tanzimCohere:
The pull request updates the `timestamp` field in the `MetricsDataBase` class and the `get_event_data` function to use the current Unix timestamp in seconds. 

## Summary
- The `timestamp` field in the `MetricsDataBase` class now has a default value of the current Unix timestamp in seconds.
- The `get_event_data` function now includes a `now_unix_seconds` variable, which stores the current Unix timestamp in seconds and is used as the `timestamp` value in the `MetricsData` object.

### DIFF
--- a/src/backend/schemas/metrics.py
+++ b/src/backend/schemas/metrics.py
@@ -40,7 +40,7 @@ class MetricsDataBase(BaseModel):
     user_id: str
     trace_id: str
     message_type: MetricsMessageType
-    timestamp: float = time.time()
+    timestamp: float
     secret: str = ""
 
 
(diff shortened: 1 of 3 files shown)


src/backend/schemas/metrics.py
```diff
@@ -36,7 +36,7 @@ class MetricsDataBase(BaseModel):
     user_id: str
     trace_id: str
     message_type: MetricsMessageType
-    timestamp: float = time.time()
+    timestamp: float
```
tanzimCohere, line 39: this likely only runs when the app is started

lusmoura (approved):

PR #451, opened 2024-07-17
[backend] file service implementation

scott-cohere:
Create a new level of abstraction between the DB/Compass and the backend feature code. We can do this by introducing a new “FileService” class implementation that backend code can call on when they want to perform operations that involve files. 

We can store files in DB/Compass as is with no dependent metadata and the backend code can simply track which files they need with a file_id. The FileService implementation can orchestrate the proper operations. For example, If we have file_ids in an entry of agent_tool_metadata, the FileService can retrieve these files and also mutate the agent_tool_metadata if needed (If we delete a file, we need to update the agent_tool_metadata, etc).

**Notes**:
This is just an initial refactor to untwist our dependencies on Fkeys for files. There will probably have to be more refactoring to introduce Compass integration. That will be a follow up to this.

DB Changes:
- Remove all details from the `Files` table, just store `file_id` and contents. Note that no other tables should have keys on this table, the file service handles syncing. This is necessary because we want the storage solutions to be agnostic to the BE code. Most notably when we integrate w/ Compass
- Conversations and Messages previously has a relationship with the Files table. Now we still store file_ids needed but through association tables, the file service again will handle syncing them

BE Changes:
- Most notably changes to router/conversation, where we replace the file_crud operations to be handled by the file service
- Changes to chat (when attaching files)

**AI Description**

This PR introduces changes to the backend, specifically enhancing the file management system. Here's a summary:

- **Database Models**:
   - `ConversationFileAssociation` and `MessageFileAssociation` tables are added to establish relationships between files, conversations, and messages.
   - The File model is updated to remove direct references to conversations and messages, now utilizing the new association tables.
- **CRUD Operations**:
   - New functions are added to handle CRUD operations for conversation and message file associations.
   - The `get_files_by_conversation_id` and `get_files_by_ids` functions are refactored to use the `FileService` class.
   - The `get_files_by_agent_id` function is introduced in the `FileService` class to retrieve files associated with a specific agent.
   - The `bulk_delete_files` function is added to the `file.py` module, allowing for efficient deletion of multiple files.
- **Router Updates**:
   - The `src/backend/routers/conversation.py` module is modified to incorporate file associations and utilize the `FileService` class for file-related operations.
   - The `getMessagesWithFiles` function is introduced to retrieve messages along with their associated files.
   - The `upload_file` and `batch_upload_file` functions are updated to use the `FileService` for file creation and association.
   - The `list_files` function now utilizes the `FileService` to retrieve files by conversation ID.
   - The `update_file` function is adjusted to use the `FileService` for file updates.
   - The `delete_file` function now invokes the `FileService` to handle file deletion and conversation association removal.
- **Tests**:
   - New test functions are added to validate the creation and deletion of conversation and message file associations.
   - The `test_list_files_by_conversation_id` function is removed, as the associated code path is no longer present.
   - The `test_list_files_by_user_id` function is updated to focus on user-specific file retrieval.
   - The `test_update_file` function is modified to reflect changes in the `update_file` router logic.
   - The `test_delete_file` function now incorporates the `FileService` for file deletion.


src/backend/routers/conversation.py
```diff
@@ -41,6 +42,8 @@
 )
 router.name = RouterName.CONVERSATION
 
+file_service = FileService()
```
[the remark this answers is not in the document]
scott-cohere: noted. here are the current validation and cleaning

https://github.com/cohere-ai/cohere-toolkit/pull/451/files#diff-4b22e9856523d0c3b06d0061724bd0962febabb12d2130f63454f1c501dda9efR52
https://github.com/cohere-ai/cohere-toolkit/pull/451/files#diff-4b22e9856523d0c3b06d0061724bd0962febabb12d2130f63454f1c501dda9efL29

@GangGreenTemperTatum is this sufficient. please lmk whats missing and ill add it.

src/backend/alembic/versions/5c3adc3d6a0a_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
```
scott-cohere, line 1: Needed to do two migrations. First to add unique constraint on convo id then the actual big changes.

src/backend/routers/conversation.py
```diff
@@ -363,32 +433,19 @@ async def batch_upload_file(
...
+        uploaded_files = await file_service.create_conversation_files(
+            session, files, user_id, conversation.id
+        )
     except Exception as e:
         raise HTTPException(
             status_code=500, detail=f"Error while uploading file(s): {e}."
         )
 
-    return uploaded_files
+    files_with_conversation_id = attach_conversation_id_to_file(
```
scott-cohere, line 445: we do this because files dont have a convo id where its stored now (goal to be agnostic to BE code). however, the endpoints and FE still needs convo id with the file. can potentially always return with file ID in the file service methods, but those methods are not always used for router logic.

src/backend/alembic/versions/5c3adc3d6a0a_.py
```diff
@@ -0,0 +1,30 @@
...
+
```
scott-cohere, line 2: note: working right now to also add a patch migration to transfer over file id, convo, message data.

src/backend/alembic/versions/5c3adc3d6a0a_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "5c3adc3d6a0a"
+down_revision: Union[str, None] = "ed17f144f4bf"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint(None, "conversations", ["id"])
```
tianjing-li, line 23: Is there a use for this unique constraint on the `id` column since this is the PrimaryKey? From what I understand, it is unique by default

src/backend/chat/custom/custom.py
```diff
@@ -8,13 +8,15 @@
...
-from backend.crud.file import get_files_by_conversation_id
+from backend.database_models.file import File
 from backend.schemas.chat import ChatMessage, ChatRole
 from backend.schemas.cohere_chat import CohereChatRequest
 from backend.schemas.tool import Tool
+from backend.services.file import FileService
 from backend.services.logger import get_logger, send_log_message
 
 logger = get_logger()
+file_service = FileService()
```
tianjing-li, line 19: nit: newline

src/backend/database_models/conversation.py
```diff
@@ -2,23 +2,41 @@
...
+    __table_args__ = (
+        UniqueConstraint("conversation_id", "file_id", name="unique_conversation_file"),
+    )
+
+
 class Conversation(Base):
     __tablename__ = "conversations"
 
-    id: Mapped[str] = mapped_column(String, default=lambda: str(uuid4()))
+    id: Mapped[str] = mapped_column(String, default=lambda: str(uuid4()), unique=True)
```
tianjing-li, line 32: Oh I see, I'm not sure why there was a overriden `id` column existing. Can you delete this column and make it use the `Base` model's `id` field?

src/backend/database_models/conversation.py
```diff
@@ -34,6 +52,13 @@ class Conversation(Base):
     def messages(self):
         return sorted(self.text_messages, key=lambda x: x.created_at)
 
+    @property
+    def file_ids(self):
```
tianjing-li, line 56: Great usage of property!

src/backend/services/conversation.py
```diff
@@ -91,6 +98,48 @@ def extract_details_from_conversation(
     return chatlog
 
 
+def getMessagesWithFiles(
```
tianjing-li, line 101: Can we make this snake_case?

src/backend/routers/conversation.py
```diff
@@ -45,15 +39,18 @@
...
 from backend.services.file import (
-    get_file_content,
+    FileService,
+    attach_conversation_id_to_files,
     validate_batch_file_size,
     validate_file,
     validate_file_size,
 )
 
+file_service = FileService()
```
tianjing-li, line 53: Suggestion: since you're going with a singleton-like implementation, rather than define the file service globally in the file that uses it, in the services/file.py can you define some method to get it?

Something like:

```
file_service = None

class FileService:
    ...

def get_file_service():
     if not file_service:
            file_service = FileService()
     
     return file_service
```

Then just import `get_file_service`, and wherever you need to use it, do 
`get_file_service().mymethod()`

tianjing-li: Ping me if you have questions about this

src/backend/services/conversation.py
```diff
@@ -91,6 +98,48 @@ def extract_details_from_conversation(
...
+        user_id (str): The user ID
+        messages (list[MessageModel]): The messages to get files for
+
+    Returns:
+        list[Message]: The messages with files
+    """
+    messages_with_file = []
+
+    for message in messages:
+        files = file_service.get_message_files(session, message.id, user_id)
```
tianjing-li, line 118: Can you make this a batch call?

src/backend/services/file.py
```diff
@@ -24,11 +31,267 @@
...
+
+        Args:
+            session (DBSessionDep): The database session
+            file_ids (list[str]): The file IDs
+            user_id (str): The user ID
+        """
+        file_crud.bulk_delete_files(session, file_ids, user_id)
+
+    def get_message_files(
+        self, session: DBSessionDep, message_id: str, user_id: str
```
tianjing-li, line 255: I would rename this to `get_files_by_message` or message_id. 

I suggest implementing a batch get as well that takes as input a List of message_ids, and using that instead where needed

tianjing-li (changes requested):
Couple requests!

src/backend/alembic/versions/5c3adc3d6a0a_.py
scott-cohere: Couple things I noticed:

- We override the base definition here for conversations https://github.com/cohere-ai/cohere-toolkit/blob/main/src/backend/database_models/conversation.py#L15, so it is not unique. The PK for conversations is id+user_id, so I suppose that was why.
- If we do not add unique constraint here we can't create a relation between convosation and conversationfileassociation. ie. ConversationFileAssociation.conversation and Conversation.conversation_file_associations

src/backend/database_models/conversation.py
scott-cohere: i believe the override is due to this table having a composite PK (id+user_id)

src/backend/database_models/conversation.py
scott-cohere: so cool that sqlalchemy orm can do this!

src/backend/services/conversation.py
scott-cohere: lmaoooo oops, old habits. 😓

src/backend/routers/conversation.py
scott-cohere: yep that makes so much more sense. thanks!

src/backend/services/file.py
scott-cohere: I'm assuming this is also related to https://github.com/cohere-ai/cohere-toolkit/pull/451#discussion_r1692215784, so I'll address these related requests here.

We want to use the method to get file(s) _per_ message. So I can't see how using a batch get files with a list of message_ids would really work here.

src/backend/tests/crud/test_conversation.py
```diff
@@ -25,6 +25,7 @@ def test_create_conversation(session, user):
     assert conversation.description == conversation_data.description
 
 
+# TODO scott: Note in PR, ask why this is allowed. Blocks relational table refactor. Needs ID to be unique.
 def test_create_conversation_same_id_different_user(session, user):
```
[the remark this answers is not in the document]
lusmoura: Unfortunately, it does =( Users are allowed to start a conversation with any `conversation_id`, similar to co.chat. We don't even enforce the conversation ID to be a UUID. So the correct primary is `conversation_id`+`user_id`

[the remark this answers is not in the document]
scott-cohere: Discussed on Slack:
- Making convo id unique
- create followup ticket to make it PK and fix all other tables referencing it,

src/backend/database_models/conversation.py
tianjing-li: Ah right, ty

src/backend/services/file.py
tianjing-li: Got it! Sgtm

tianjing-li (approved):

PR #461, opened 2024-07-18
metrics: chat events

mtanzim:
This PR introduces changes to the backend, specifically enhancing the chat functionality and associated tests. 

**Backend Changes:**
- Added `request` as a parameter to the `generate_chat_response` and `generate_chat_stream` functions in `chat.py`.
- Modified the `generate_conversation_title` function in `conversation.py` to include `request` as a parameter.
- Updated the `run-tests` command in `Makefile` to build and run tests for the backend.
- Introduced the `add_model_to_request_state` function in `utils.py` to set the model attribute in the request state.
- Removed the comment indicating that the `CHAT_API_SUCCESS` and `CHAT_API_FAIL` event types are implemented and have tests.
- Created a new `MetricsChat` class in `metrics.py` to store chat-related metrics.
- Modified the `MetricsData` class in `metrics.py` to include the `error` attribute.
- Added the `report_streaming_event` function in `services/chat.py` to report streaming events.
- Updated the `generate_chat_response` and `generate_chat_stream` functions in `services/chat.py` to include `request` as a parameter.
- Modified the `generate_chat_stream` function in `services/chat.py` to report streaming events.
- Refactored the `MetricsMiddleware` class in `services/metrics.py` to confirm the environment, initialize request state, and send signals.
- Introduced the `get_agent` function in `services/metrics.py` to retrieve the agent from the request state.
- Added the `report_streaming_event` function in `services/metrics.py` to handle streaming events.

**Test Changes:**
- Added a new test case, `test_streaming_new_chat_metrics_with_agent`, in `test_chat.py` to verify chat streaming metrics with an agent.
- Updated the `test_streaming_new_chat_with_agent` test case in `test_chat.py` to include the agent ID in the JSON data.
- Modified the `test_streaming_new_chat_with_agent_existing_conversation` and `test_streaming_chat_with_existing_conversation_from_other_agent` test cases in `test_chat.py` to include the agent ID in the JSON data.
- Removed the `test_streaming_fail_chat_missing_message` test case, which was previously skipped due to the absence of the Cohere API key.


src/backend/services/metrics.py
```diff
@@ -40,48 +41,69 @@
...
 
+        self.confirm_env()
+        self.init_req_state(request)
+
+        start_time = time.perf_counter()
+        response = await call_next(request)
+        duration_ms = time.perf_counter() - start_time
+        self.send_signal(request, response, duration_ms)
+
+        # self.process_signal_queue(request, response)
```
mtanzim, line 54: remove

src/backend/tests/routers/test_chat.py
```diff
@@ -264,7 +314,7 @@ def test_default_chat_missing_deployment_name(
 
 
 @pytest.mark.skipif(not is_cohere_env_set, reason="Cohere API key not set")
-def test_streaming_fail_chat_missing_message(
+def test_streaming_fail_chat_missing_message_logs_metric(
```
mtanzim, line 317: TODO: revert this

src/backend/schemas/metrics.py
```diff
@@ -20,13 +20,11 @@ class MetricsMessageType(str, Enum):
     ASSISTANT_UPDATED = "assistant_updated"
     ASSISTANT_DELETED = "assistant_deleted"
     ASSISTANT_ACCESSED = "assistant_accessed"
-    # chat: implemented
+    # pending implementation
```
mtanzim, line 23: TODO: update

src/backend/schemas/metrics.py
```diff
@@ -61,14 +59,20 @@ class MetricsAgent(BaseModel):
     description: str | None
 
 
+class MetricsChat(BaseModel):
+    input_nb_tokens: int
+    output_nb_tokens: int
+    search_units: int
+    model: str
+    assistant_id: str | None
```
mtanzim, line 67: align on if we care about messages without assistants, and/or we should simply latch on the default assistant id

src/backend/services/metrics.py
```diff
@@ -40,48 +41,69 @@
...
         request.state.user = None
         request.state.event_type = None
 
-        start_time = time.perf_counter()
-        response = await call_next(request)
-        duration_ms = time.perf_counter() - start_time
-        data = self.get_event_data(request.scope, response, request, duration_ms)
-        signal = preprocess_event_data(data)
-        should_send_event = request.state.event_type and data and signal
+    def process_signal_queue(self, request: Request, response: Response) -> None:
```
mtanzim, line 65: todo: remove, not using

src/backend/services/metrics.py
```diff
@@ -192,13 +175,86 @@ def log_signal_curl(signal: MetricsSignal) -> None:
...
+                input_nb_tokens=input_tokens,
+                output_nb_tokens=output_tokens,
+                search_units=search_units,
+                model=model,
+                assistant_id=agent_id,
+            )
+        else:
+            import pdb
+
+            pdb.set_trace()
```
mtanzim, line 232: argh: remove

Makefile
```diff
@@ -12,6 +12,8 @@ run-agent-router-tests:
 	docker compose run --build backend poetry run pytest src/backend/tests/routers/test_agent.py
 run-user-router-tests:
 	docker compose run --build backend poetry run pytest src/backend/tests/routers/test_user.py
+run-chat-router-tests:
```
tianjing-li, line 15: Is there a reason we are adding new make commands instead of using the `run-tests` one with an arg?

src/backend/schemas/metrics.py
tianjing-li: It would still be possible to create /chat requests without an Agent and just a model deployment, so probably good to track those too

Makefile
mtanzim: will remove, just convenient for local dev

tianjing-li: Thanks yeah, let's just remove those to avoid cluttering the makefile too much

tianjing-li (approved):
lgtm, just small comment (see slack)

src/backend/schemas/metrics.py
mtanzim: aligned on passing in default assistant to events when the agent is not specified

PR #464, opened 2024-07-19
feat(toolkit): do not override google drive files on new selection

knajjars:
**AI Description**

This pull request updates the `CreateAgent` and `UpdateAgent` components in the `src/interfaces/coral_web/src/components/Agents` directory. 

**Summary**
- The PR introduces changes to the way artifacts are handled when the tool name is `TOOL_GOOGLE_DRIVE_ID`. 
- It ensures that duplicate artifacts are removed by using the `uniqBy` function from Lodash. 
- The `tools_metadata` array is updated to include only unique artifacts for `TOOL_GOOGLE_DRIVE_ID`, while keeping other tool artifacts unchanged. 

**Code Changes**
- The `uniqBy` function from Lodash is imported in both `CreateAgent.tsx` and `UpdateAgent.tsx`.
- In `CreateAgent.tsx`, when handling data from Google Drive, the code now constructs an `updatedArtifacts` array, which includes existing artifacts for `TOOL_GOOGLE_DRIVE_ID` and new artifacts from the `data.docs` array. 
- The `tools_metadata` array is then updated with a new object containing `TOOL_GOOGLE_DRIVE_ID` and the unique `updatedArtifacts`. 
- Similar changes are made in `UpdateAgent.tsx`, where the `updatedArtifacts` array is constructed and used to update the `artifacts` property of the `updateGoogleDriveTool` object.

### DIFF
--- a/src/interfaces/coral_web/src/components/Agents/CreateAgent.tsx
+++ b/src/interfaces/coral_web/src/components/Agents/CreateAgent.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { useLocalStorageValue } from '@react-hookz/web';
+import { uniqBy } from 'lodash';
 import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import React, { useContext, useEffect, useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


src/interfaces/coral_web/src/components/Agents/CreateAgent.tsx
```diff
@@ -54,26 +55,35 @@ export const CreateAgent: React.FC = () => {
...
-                    id: doc.id,
-                    name: doc.name,
-                    type: doc.type,
-                    url: doc.url,
-                  } as GoogleDriveToolArtifact)
-              ),
-            },
+      setFields((prev) => {
+        const updatedArtifacts = [
+          ...(prev.tools_metadata?.find((tool) => tool.tool_name === TOOL_GOOGLE_DRIVE_ID)
```
tomtobac, line 60: Can we extract this function into a variable? so we can use it later, like:
```tsx
const hasGoogleDriveTool = prev.tools_metadata?.find((tool) => tool.tool_name === TOOL_GOOGLE_DRIVE_ID;
```

tomtobac (approved):

src/interfaces/coral_web/src/components/Agents/CreateAgent.tsx
knajjars: I don't find very useful to do this abstraction, in this file it is referenced only here

knajjars: the other place where we iterate over tools, is to select all, except google drive tool, so this will also not be very helpful

PR #465, opened 2024-07-19
feat(toolkit): use google drive icon svg

knajjars:
**AI Description**

This pull request introduces a new `svg` prop to the `DocumentIcon` component and updates the `CitationDocumentHeader` component to use it. The `svg` prop is used to render an SVG icon for a given tool ID, allowing for more flexible and customizable icons.

- The `DocumentIcon` component now accepts an `svg` prop, which is used to render an SVG icon. The `url` prop is also made optional.
- The `CitationDocumentHeader` component now uses the `svg` prop of the `DocumentIcon` component to render the icon for a tool. It retrieves the SVG content based on the tool ID from the `TOOL_ID_TO_ICON` object in `constants.ts`.
- A new `GoogleDriveIconSVG` constant is added to `constants.ts`, which contains the SVG content for the Google Drive icon.
- The `getHostname` function in `DocumentIcon.tsx` now accepts an optional `url` parameter to handle cases where the URL is not provided.


src/interfaces/coral_web/src/constants.ts
```diff
@@ -67,6 +70,9 @@ export const TOOL_ID_TO_DISPLAY_INFO: { [id: string]: { icon: IconName } } = {
   [TOOL_WIKIPEDIA_ID]: { icon: 'web' },
   [TOOL_SEARCH_FILE_ID]: { icon: 'search' },
 };
+export const TOOL_ID_TO_ICON: { [id: string]: StaticImport } = {
```
tomtobac, line 73: there is not need to specify the import type here since its inherit

tomtobac (approved):

src/interfaces/coral_web/src/constants.ts
knajjars: .svg imports are inferred as "any" in nextjs

PR #481, opened 2024-07-22
metrics: rerank events

mtanzim:
since ai desc not working:

## Key Changes

- Added rerank events tracking and a test to ensure happy path event (leverages conversations:search endpoint)
- Used decorators for the implementation, but relying quite heavily on request state
- Refactored metrics module and added ai gen docstrings; hoping it'll help us keep improving/refactoring as we add adjacent features such as requests contexts;
- Also added back decorators for chat streams for uniformity

Please review in detail:
- decorator logic and placement
- organization and design decisions: namespaces using classes with static methods, usage of request state etc
- suggestions for additional tests


src/backend/model_deployments/cohere_platform.py
```diff
@@ -90,11 +95,11 @@ async def invoke_chat_stream(
             )
             yield to_dict(event)
 
+    @collect_metrics_rerank
```
lusmoura, line 98: We need the decorator in all model deployments for consistency

src/backend/model_deployments/cohere_platform.py
```diff
@@ -90,11 +95,11 @@ async def invoke_chat_stream(
...
     async def invoke_rerank(
         self, query: str, documents: List[Dict[str, Any]], **kwargs: Any
     ) -> Any:
-        response = self.client.rerank(
-            query=query, documents=documents, model=DEFAULT_RERANK_MODEL
-        )
-
+        model = DEFAULT_RERANK_MODEL
+        add_rerank_model_to_request_state(model, **kwargs)
```
lusmoura, line 103: I'm not sure if this is the best place to add the rerank model to the state since we would need to add the same snippet to all deployments. Wdyt about defaulting the model in the metrics service? If we ever decide to support model selection, we can follow up with this change.

src/backend/model_deployments/cohere_platform.py
tanzimCohere: addressed

src/backend/model_deployments/cohere_platform.py
tanzimCohere: addressed

scott-cohere (approved):

PR #483, opened 2024-07-23
Pull settings from YAML files and the ENV vars 

BeatrixCohere:
- Support yaml config 
- Also looks at env vars as priority to not break current set up 
- Only for backend currently 
- Secrets split into another file 

**AI Description**

This PR introduces a new `Settings` class to handle environment variables and configuration settings. The `Settings` class is designed to read environment variables from a `.env` file and provide a centralized way to access these variables and configuration settings in the codebase.

- The `Settings` class is defined in <co: 12>`src/backend/config/settings.py`. It uses the `pydantic_settings` module to handle the loading of environment variables and configuration files.
- The `settings_customise_sources` class method is defined to specify the priority order of the settings sources: `env_settings`, `dotenv_settings`, `configuration.yaml`, `secrets.yaml`, `file_secret_settings`, and `init_settings`.
- The `Settings` class contains various nested classes, such as `AuthSettings`, `FeatureFlags`, `PythonToolSettings`, `DatabaseSettings`, and specific deployment settings classes (`SageMakerSettings`, `AzureSettings`, etc.). These nested classes provide structured access to related settings.
- The `get_available_tools` function in `src/backend/config/tools.py` now uses the `Settings` class to access the `use_langchain_tools` and `use_community_tools` settings.
- The `get_available_deployments` function in `src/backend/config/deployments.py` now uses the `Settings` class to filter the `ALL_MODEL_DEPLOYMENTS` dictionary based on the `enabled_deployments` setting.
- The `get_default_deployment` function in `src/backend/config/deployments.py` now uses the `Settings` class to determine the default deployment.
- The `get_model_config_var` function in `src/backend/model_deployments/utils.py` now accepts a default value as a parameter. If the config value is not found, the default value is returned instead of raising a `ValueError`.
- The `load_dotenv` function is called in `src/backend/config/auth.py` and `src/backend/database_models/database.py` to load environment variables from the `.env` file.
- The `MIGRATE_TOKEN` variable is now accessed through the `Settings` class in `src/backend/config/auth.py` and `src/backend/database_models/database.py`.
- The `ENABLED_AUTH_STRATEGIES` list in `src/backend/config/auth.py` is now populated based on the `Settings` class.
- The `auth_secret` variable in `src/backend/main.py` is now retrieved from the `Settings` class.
- The `auth_key` variable in `src/backend/services/auth/crypto.py` is now retrieved from the `Settings` class.
- The `secret_key` variable in `src/backend/services/auth/jwt.py` is now retrieved from the `Settings` class.
- The `GoogleOAuth` and `OpenIDConnect` classes in `src/backend/services/auth/strategies/google_oauth.py` and `src/backend/services/auth/strategies/oidc.py` now use the `Settings` class to access the required settings.
- The `redirect_uri` variable in `src/backend/routers/auth.py` is now retrieved from the `Settings` class.
- The `use_langchain` variable in `src/backend/routers/chat.py` is now retrieved from the `Settings` class.
- The `list_experimental_features` function in `src/backend/routers/experimental_features.py` now uses the `Settings` class to access the `use_experimental_langchain` and `use_agents_view` settings.
- The `WOLFRAM_ALPHA_APP_ID` variable in `.env-template` has been updated to `WOLFRAM_APP_ID`.
- The `.gitignore` file has been updated to include `src/backend/config/configuration.yaml` and `src/backend/config/secrets.yaml`.
- The `docker-compose.yml` file has been updated to include mounts for the new configuration and secrets files.
- The `get_deployment` function in `src/backend/chat/custom/utils.py` has been modified to use the `get_default_deployment` function and updated to return the default deployment if available.
- The `COHERE_API_ENV_VAR` import in `src/backend/compass_sdk/constants.py` has been removed.
- The `cohere_api_key` variable in `src/backend/compass_sdk/__init__.py` now uses the `CohereDeployment.api_key` attribute.


src/backend/config/auth.py
```diff
@@ -5,14 +5,23 @@
...
 load_dotenv()
 
+auth_map = {
+    "basic": BasicAuthentication,
+    "google_oauth": GoogleOAuth,
+    "oidc": OpenIDConnect,
+}
+
 SKIP_AUTH = os.getenv("SKIP_AUTH", None)
 # Add Auth strategy classes here to enable them
```
lusmoura, line 20: nit: remove this comment

src/backend/config/deployments.py
```diff
@@ -26,39 +25,44 @@ class ModelDeploymentName(StrEnum):
...
 
 
-use_community_features = bool(strtobool(os.getenv("USE_COMMUNITY_FEATURES", "false")))
-
+use_community_features = Settings().feature_flags.use_community_features
 
+# TODO names in the map below should not be the display names but ids
 ALL_MODEL_DEPLOYMENTS = {
     ModelDeploymentName.CoherePlatform: Deployment(
+        id="cohere_platform",
```
lusmoura, line 33: Should we have an enum with the ids? It might be easier to find all existing tools when deciding which ones to enable

src/backend/config/tools.py
```diff
@@ -212,6 +209,10 @@ def get_available_tools() -> dict[ToolName, dict]:
         # Retrieve name
         tool.name = tool.implementation.NAME
 
+    enabled_tools = Settings().tools.enabled_tools
```
lusmoura, line 212: I think we should add docs explaining the difference between configuring the yaml and the community package

lusmoura (approved):

src/backend/config/deployments.py
BeatrixCohere: Yeah, I added a TODO we need the key of the map to also be the ID not the display name

src/backend/config/tools.py
BeatrixCohere: Will add more docs around this once it is more tested - for now all the env things should still work

src/backend/config/configuration.yaml
```diff
@@ -0,0 +1,44 @@
+
```
giannis2two, line 1: nit

src/backend/config/configuration.template.yaml
```diff
@@ -0,0 +1,51 @@
+
```
giannis2two, line 1: nit

src/backend/config/settings.py
```diff
@@ -0,0 +1,264 @@
...
+    enabled_deployments: Optional[List[str]]
+
+    sagemaker: Optional[SageMakerSettings]
+    azure: Optional[AzureSettings]
+    cohere_platform: Optional[CoherePlatformSettings]
+    single_container: Optional[SingleContainerSettings]
+    bedrock: Optional[BedrockSettings]
+
+
+class Settings(BaseSettings, case_sensitive=False):
```
giannis2two, line 229: 🔥

giannis2two, line 229: Should we initialize a settings instance once and put it in `backend/config/__init__.py` with @lru_cache something like
```
@lru_cache(maxsize=1)
def settings() -> Settings:
    try:
        return Settings()
    except Exception as e:
        # handle
```

Prevents from initializing many Settings instances

src/backend/tools/base.py
```diff
@@ -47,9 +48,10 @@ class BaseToolAuthentication:
     """
 
     def __init__(self, *args, **kwargs):
-        self.BACKEND_HOST = os.getenv("NEXT_PUBLIC_API_HOSTNAME")
-        self.FRONTEND_HOST = os.getenv("FRONTEND_HOSTNAME")
-        self.AUTH_SECRET_KEY = os.getenv("AUTH_SECRET_KEY")
+        auth_config = Settings().auth
```
giannis2two, line 51: Maybe keep consistent and remove this? Keep Settings().auth.* for each

.env-template
```diff
@@ -9,7 +9,7 @@ MIGRATE_TOKEN=<MIGRATE_TOKEN>
 # TOOLS
```
giannis2two, line 9: Is an env template needed now?

giannis2two (approved):
Awesome PR!!

.env-template
BeatrixCohere: Going to keep env's around also in the docs until this is tested more

src/backend/config/settings.py
BeatrixCohere: Added, is the error handling necessary? I find the errors from settings quite good

PR #500, opened 2024-07-25
[1/N] Logs: add context

lusmoura:
1/N - The first step to improving our logs is to enhance our codebase by adding a Context type to wrap all relevant fields. I couldn't split this into smaller PRs, so sorry in advance if it's too difficult to review. 😅 

**AI Description**

This PR introduces a new `Context class and makes changes to the code to use it. The `Context` class is defined in `src/backend/schemas/context.py`. It holds information such as `request`, `response`, `trace_id`, `user_id`, `event_type`, `user`, `agent`, etc. 

A new middleware, `ContextMiddleware`, is added in `src/backend/services/context.py`. It creates a new context for each request, setting the request, response, trace ID, user ID, agent ID, and deployment name. It also clears the context after the request is complete. 

The `get_context` function is added in `src/backend/services/context.py` to retrieve the context from the request scope. 

The code in the following files is updated to use the `Context` class and related functions:
- `src/backend/chat/collate.py`
- `src/backend/chat/custom/custom.py`
- `src/backend/chat/custom/utils.py`
- `src/backend/main.py`
- `src/backend/model_deployments/azure.py`
- `src/backend/model_deployments/base.py`
- `src/backend/model_deployments/bedrock.py`
- `src/backend/model_deployments/cohere_platform.py`
- `src/backend/model_deployments/sagemaker.py`
- `src/backend/model_deployments/single_container.py`
- `src/backend/model_deployments/utils.py`
- `src/backend/routers/agent.py`
- `src/backend/routers/auth.py`
- `src/backend/routers/chat.py`
- `src/backend/routers/conversation.py`
- `src/backend/routers/deployment.py`
- `src/backend/routers/experimental_features.py`
- `src/backend/routers/snapshot.py`
- `src/backend/routers/tool.py`
- `src/backend/routers/user.py`
- `src/backend/services/conversation.py`
- `src/backend/services/logger.py`
- `src/backend/services/metrics.py`
- `src/backend/services/utils.py`
- `src/backend/tools/tavily.py`


src/backend/services/metrics.py
```diff
@@ -256,38 +234,27 @@ def log_signal_curl(signal: MetricsSignal) -> None:
...
+        event: dict[str, Any], ctx: Context, **kwargs: Any
+    ) -> None:
+        start_time = None
         try:
-            request = kwargs.get("request", None)
-            if not request:
-                raise ValueError("request not set")
             event_type = event["event_type"]
             if event_type == StreamEvent.STREAM_START:
-                request.state.stream_start = time.perf_counter()
```
mtanzim, line 276: removing this will likely break the duration feature, the `start_time` variable is destroyed as the function exits and not persisted across the duration of the stream

src/backend/schemas/context.py
```diff
@@ -0,0 +1,154 @@
...
+from backend.services.utils import get_deployment_config
+
+
+class Context(BaseModel):
+    request: Optional[dict] = {}
+    response: Optional[dict] = {}
+    receive: Optional[dict] = {}
+    trace_id: str = "default"
+    user_id: str = "default"
+    event_type: MetricsMessageType = None
```
tanzimCohere, line 20: can this default to `MetricsMessageType.UNKNOWN_SIGNAL`? it'll limit the type to just MetricsMessageType

tanzimCohere: can we also move it down with `metrics_user`
`metrics_agent`

src/backend/schemas/context.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if user:
+            self.metrics_user = MetricsUser(
+                id=user.id, email=user.email, fullname=user.fullname
+            )
+            self.user = user
+
+        return self
+
+    def with_agent(self, agent: Agent | None) -> "Context":
```
mtanzim, line 73: should we use the same function to manage agent and metrics_agent? there also seem to be a short circuit when agent is not defined? should the function exit before defining `self.metrics_agent`?

src/backend/services/metrics.py
```diff
@@ -62,25 +65,14 @@ class MetricsMiddleware(BaseHTTPMiddleware):
     """
 
     async def dispatch(self, request: Request, call_next: Callable):
-
-        self._confirm_env()
```
mtanzim, line 66: why did we remove this? It'll help identify metrics config errors easily

src/backend/services/metrics.py
mtanzim: One solution can be to track stream start in the context, and then pulling it out to calcualte duration ms, similar to how we were leveraging `request.state`

src/backend/routers/chat.py
```diff
@@ -49,57 +45,47 @@ async def chat_stream(
...
     agent_id = chat_request.agent_id
+    ctx.with_agent_id(agent_id)
+
+    agent = DEFAULT_METRICS_AGENT
     if agent_id:
         agent = agent_crud.get_agent_by_id(session, agent_id)
-        add_agent_to_request_state(request, agent)
-    else:
-        add_default_agent_to_request_state(request)
+    ctx.with_agent(agent)
```
mtanzim, line 61: `Agent` and `MetricsAgent` are different types and don't have a subtype r/s; using them in this way can lead to bugs

mtanzim (changes requested):
Fantastic improvements! Can you double check comments around the following key points:
- ~~Distinguishing b/w Agent/MetricsAgent and ensuring the fallback for MetricsAgent is being triggered correctly. Unfortunately there are no tests for it iirc 😞  , could be a good chance to add one~~
- ^ never mind, there is actually test coverage 🎉 would still recommend _Distinguishing b/w Agent/MetricsAgent_ 
- Ensuring duration is calculation is working correctly for streaming chat events; similar issue with tests as above

src/backend/schemas/context.py
mtanzim: actually, I take this back, the check is currently `event_type is None`

src/backend/services/context.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ContextMiddleware:
+    def __init__(self, app: ASGIApp):
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        trace_id = str(uuid.uuid4())
+
+        if scope["type"] not in ["http", "websocket"]:
```
scott-cohere, line 17: can we remove websocket? we don't use it in toolkit right?

scott-cohere (approved):
this is such a big dev experience improvement. thanks!

what are the follow up items for this since its 1/n

src/backend/services/metrics.py
lusmoura: Oh I didn't know about the variable was destroyed 😅  good catch, I'll add it to the context!

src/backend/schemas/context.py
lusmoura: Sure, I can split this into two functions np

src/backend/services/metrics.py
lusmoura: My bad, I removed to test locally

lusmoura:
> this is such a big dev experience improvement. thanks!
> 
> what are the follow up items for this since its 1/n

Follow ups: introduce structured logs, remove unused logs and improve quality/coverage

src/backend/schemas/context.py
lusmoura: Before my changes, we checked whether the agent existed, and if so, we added the metrics agent to the request state. Now, we don't check that before trying to add it to the context, so there's a chance we'll call `with_agent` with `agent = None`. In that case, we don't have any information to extract from the agent, and we can't define metrics_agent.

Is there a case in which we want to have only `agent` without having `metrics_agent`?

mtanzim: I'd suggest separating the logic, re: https://github.com/cohere-ai/cohere-toolkit/pull/500#discussion_r1691885090

One approach would be:
-  create a new function: `with_metrics_agent(a: MetricsAgent)`
- When we want to track the metrics agent, we use the above function
- If agent doesn't exist, it's an error case, handle as needed
- If default agent is used (through new default agent endpoint), we call `with_metrics_agent(DEFAULT_METRICS_AGENT)`

lusmoura: Oooooh okay, I get it now. Lemme make the changes

src/backend/schemas/metrics.py
```diff
@@ -98,3 +98,16 @@ class MetricsSignal(BaseModel):
     preamble="",
     description="default",
 )
+
+
+def agent_to_metrics_agent(agent: Any) -> MetricsAgent:
```
tanzimCohere, line 103: can we type this as `Agent`?

mtanzim (approved):
awesome work 🚀

PR #501, opened 2024-07-25
infra: Fix the smoke test when new releases are published

malexw:
There was a syntax error stopping Docker from running the container we just built. Also, I discovered that curl now has a built-in option for exiting with non-zero status codes when requests fail, which simplifies the test script considerably.

**AI Description**

The PR simplifies the Docker image testing process in the `build_and_push_docker.yml` workflow file. 

### Changes
- The `docker run` command now uses `${{ env.REGISTRY }}` and `${{ env.IMAGE_NAME }}` for improved flexibility and consistency.
- Reduced the sleep time from 30 to 20 seconds for quicker feedback.
- Replaced the `status_code` check with `curl --fail-with-body`, removing the need for an explicit status code comparison and simplifying error handling.


tianjing-li (approved):
lgtm

PR #518, opened 2024-07-29
Context: fix no auth

lusmoura:

**AI Description**

This pull request updates the `context.py` file in the backend services to handle authentication for endpoints that do not require it.

## Summary
The changes in this PR ensure that the `user_id` is handled correctly, even when it is not present in the request headers. This is important because not all endpoints require authentication, and the code needs to gracefully handle such cases.

## Changes
- Imports the `DEFAULT_USER_ID` from the `backend.schemas.user` module.
- Moves the `request = Request(scope)` line after setting the deployment name in the request headers.
- Adds a comment explaining the need to handle the case where `user_id` is not present in the request headers.
- Wraps the `user_id = get_header_user_id(request)` line in a try-except block to handle potential errors when retrieving the `user_id`.
- Sets the `user_id` to `DEFAULT_USER_ID` if an exception occurs during the `get_header_user_id(request)` call.


giannis2two (approved):

src/backend/services/context.py
```diff
@@ -28,8 +29,12 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send):
         request = Request(scope)
         context.with_deployment_name(request.headers.get("Deployment-Name", ""))
 
-        request = Request(scope)
-        user_id = get_header_user_id(request)
+        # Not all endpoints require authentication, so we need to handle the case where the user_id is not present
+        try:
+            user_id = get_header_user_id(request)
```
BeatrixCohere, line 34: Would it be better to check if authorisation is none in the func and then default? Worried this might hide another issue

lusmoura: I feel like these workaround solutions will hide other issues until we handle auth in all endpoints. I can check if auth is none for now, but I'll try to come up with something better later.

giannis2two (approved):

PR #528, opened 2024-07-30
[5/N] Logs: bind context to logger + unify loggers

lusmoura:
- Instantiate only one logger per app instead of one per file.
- Bind context, user_id and conversation_id to logger in the middleware so that all logs contain this information

**AI Description**

This PR updates the logger setup and usage across the backend code. Previously, the `get_logger` function was used to initialize a logger instance, but now a `LoggerFactory` class has been introduced, providing a more flexible and structured approach to logger creation.

- The `get_logger` function has been removed, and the `backend.services.logger.utils` module now exports a `logger` instance directly.
- The `LoggerFactory` class in `src/backend/services/logger/utils.py` encapsulates the logger creation logic. It allows for centralized control over the logger's strategy, level, and renderer.
- Various backend modules have been updated to import and use the `logger` instance directly, removing the need for explicit initialization.
- The `LoggingMiddleware` class in `src/backend/services/logger/middleware.py` now includes context binding to the logger, enhancing log messages with user and trace information.


src/backend/services/logger/utils.py
```diff
@@ -2,17 +2,28 @@
...
+        strategy = Settings().logger.strategy
+        level_str = Settings().logger.level
+        renderer = Settings().logger.renderer
+
+        level = getattr(logging, level_str.upper(), None)
+
+        if strategy == "structlog":
+            return StructuredLogging(level, renderer)
+        else:
+            raise ValueError(f"Invalid logger strategy: {strategy}")
```
BeatrixCohere, line 26: Should we default?

src/backend/services/logger/middleware.py
```diff
@@ -16,6 +14,13 @@ async def dispatch(self, request: Request, call_next):
 
         ctx = get_context(request)
```
BeatrixCohere, line 15: Is context always not None?

BeatrixCohere (approved):

arksenu (approved):

arksenu (approved):

src/backend/services/logger/middleware.py
lusmoura: Yep, it's either the existing context or a newly instantiated one, never None

PR #543, opened 2024-07-31
Migrate auth to cookies instead of localStorage

knajjars:
This PR changes the authToken to be stored in `cookies` instead of `localStorage`.
This is particularly helpful for nextjs to authenticate on the server-side and perform prefetch queries for faster response times. Also, allows for a faster redirect on unauthenticated sessions.

To test you have to start the `assistants_web` UI on port 4000 and update `.env` and `env.development` to localhost:4000 as well, since OIDC redirects to this port

**AI Description**

This PR introduces authentication to the web interface. 

**Changes:**
- Adds `next-client-cookies` package and its dependencies.
- Updates `package.json` to include `next-client-cookies`.
- Adds `COOKIE_KEYS` import and `cookies` from `'next/headers'` in `layout.tsx`.
- Adds `AgentLeftPanel` and `MobileHeader` components to `layout.tsx`.
- Removes `ProtectedPage` component and replaces it with authentication logic.
- Redirects to the login page if no auth token is found in the cookie store.
- Updates `LayoutProviders` to accept an optional `authToken` prop.
- Adds `clearAuthToken` function to `actions.server.ts`.
- Adds `setAuthToken` function to `actions.server.ts`.
- Updates `useSession` hook to use `COOKIE_KEYS` and `useCookies` hook.
- Updates `useSession` to throw an error on invalid login and await `setAuthToken`.
- Updates `useSession` to use `clearAuthToken` for logout.
- Updates `useSession` to use `setAuthToken` for successful login and OAuth.


src/interfaces/assistants_web/src/hooks/session.ts
```diff
@@ -55,15 +51,18 @@ export const useSession = () => {
...
   const loginMutation = useMutation<JWTResponse | null, ApiError, LoginParams>({
     mutationFn: (params) => cohereClient.login(params),
-    onSuccess: (data: JWTResponse | null) => {
-      setAuthToken(data?.token);
+    onSuccess: async (data: JWTResponse | null) => {
+      if (!data) {
+        return new Promise((_, reject) => reject(new Error('Invalid login')));
+      }
+      await setAuthToken(data.token);
       return new Promise((resolve) => resolve(data?.token));
```
malexw, line 59: With this function now being async, I think we can change the `return` statement to just be
```
return data?.token
```

(This is also true for the other 2 functions here that return Promise-wrapped values)

tomtobac (approved):
🏆

PR #556, opened 2024-08-01
Support PDF for file scraping 

BeatrixCohere:
Check the content type and if it is PDF read the PDF instead of HTML 

**AI Description**

The `call` function in `web_scrape.py` has been updated to handle PDF content in addition to HTML. 

Previously, the function only processed HTML content, extracting text using BeautifulSoup. Now, it can also handle PDF content by checking the "content-type" header in the response. If the content type is "application/pdf", the function uses `PdfReader` to extract text from each page of the PDF. For HTML content ("text/html"), it continues to use BeautifulSoup as before. If the content type is neither PDF nor HTML, a `ValueError` is raised.

**Code Changes:**
- New imports: `io` and `PdfReader` from `pypdf`.
- The `content_type` is checked using `response.headers.get("content-type")`.
- Conditional blocks added to handle PDF and HTML content types.
- For PDF, the `PdfReader` class is used to extract text from each page.
- The extracted text is accumulated in the `text` variable and returned in the same format as before.
- A `ValueError` is raised if the content type is unsupported.


src/backend/tools/web_scrape.py
```diff
@@ -28,14 +30,34 @@ async def call(self, parameters: dict, **kwargs: Any) -> List[Dict[str, Any]]:
...
-        return [
-            (
-                {
-                    "text": text,
-                    "url": url,
-                }
-            )
-        ]
+        content_type = response.headers.get("content-type")
+        if "application/pdf" in content_type:
```
lusmoura, line 34: You can reuse the read_pdf function in the file service here

BeatrixCohere: Done

lusmoura (approved):

PR #566, opened 2024-08-01
feat: scim

raphael-cohere:
This PR adds SCIM functionality to the project. It allows an organisation to sync their users and groups from their idp.

Allows synchronisation of users and groups from Okta to toolkit. While SCIM is an open standard, this implementation was only tested with Okta. If we want to add more providers we would need to retest, but as long as they are following the spec it should be fine (similar to OIDC).

## General Notes
- this only works when OIDC is enabled. We match the users from the OIDC request to the users that were synchronised. (via external id)
- this feature should only be activated if there were no users created so far, otherwise there could be side effects

## Resources
- okta reference: https://developer.okta.com/docs/reference/scim/scim-20/#scim-user-operations
- flask sample app: https://github.com/oktadev/okta-scim-flask-example/tree/master
- original spec: https://www.rfc-editor.org/rfc/rfc7643.html#page-

## PR Template

**AI Description**

This PR introduces the System for Cross-domain Identity Management (SCIM) to the backend. SCIM is a standard for automating the exchange of user identity information between identity domains, such as between an enterprise identity management system and a service provider.

The changes include:
- Adding SCIM configuration to the `.env-template` file.
- Creating a new `SCIMAuthValidation` class in `request_validators.py` for SCIM authentication.
- Adding SCIM-related imports and a new `SCIM` router in `routers.py`.
- Adding SCIM-related fields to the `secrets.template.yaml` and `settings.py` files.
- Creating new CRUD operations for groups and users in `group.py` and `user.py`, respectively.
- Creating new SCIM exception handling and router in `scim.py`.
- Adding SCIM-related imports and classes in `schemas/scim.py`.
- Adding a `ScimAuthValidation` class in `request_validators.py` for SCIM authentication.
- Updating the `get_or_create_user` function in `utils.py` to use the `get_user_by_external_id` function.
- Adding SCIM-related tests in `test_scim.py`.
- Adding SCIM-related configuration to the `secrets.yaml` file.


src/backend/services/logger/middleware.py
```diff
@@ -19,10 +21,12 @@ async def dispatch(self, request: Request, call_next):
         logger.info(
             event="Request",
             method=request.method,
+            body=request_body,
             path=request.url.path,
             status_code=response.status_code,
             duration=time.time() - start_time,
-            ctx=ctx,
+            response=response,
+            # ctx=ctx,
```
mtanzim, line 29: revert this before merge

src/backend/services/logger/middleware.py
```diff
@@ -11,6 +12,7 @@
 
 class LoggingMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next):
+        request_body = await request.body()
```
mtanzim, line 15: this too, might cause perf issues parsing body multiple times, was using for debug

src/backend/services/context.py
```diff
@@ -31,7 +33,8 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send):
         request = Request(scope)
         context.with_deployment_name(request.headers.get("Deployment-Name", ""))
 
-        if has_header_user_id(request):
+        ignore_path = any(request.url.path.startswith(p) for p in _ignore_paths)
+        if not ignore_path and has_header_user_id(request):
```
raphael-cohere, line 37: It would be nice to find a better solution for this. Basically the scim routes are being authenticated by setting the header to : "Authorisation: ....". But the current code always assume there is a jwt token in the auth header and fails to parse.

src/backend/services/auth/utils.py
```diff
@@ -37,12 +37,16 @@ def get_or_create_user(session: Session, token_user: dict[str, str]) -> dict:
     """
     email = token_user.get("email")
     fullname = token_user.get("name")
+    external_id = token_user.get("sub")
```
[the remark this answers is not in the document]
raphael-cohere: yes this is in the OIDC spec

src/backend/services/context.py
raphael-cohere: From beatrix:

```
Yeah, I think the optimal thing would be:
Try and get JWT token
success use as user_id
fails continue
Try and get user id from header
success use as user_id
fails continue
Is SCIM request
success use "scim" as user_id
fails continue
Use default user id
```

src/backend/routers/scim.py
```diff
@@ -0,0 +1,290 @@
...
+    ListGroupResponse,
+    ListUserResponse,
+    PatchGroup,
+    PatchUser,
+    UpdateUser,
+    User,
+)
+
+
+def _check_basic_auth(
```
raphael-cohere, line 31: maybe move this somewhere to make it reusable?

src/backend/services/auth/utils.py
```diff
@@ -37,12 +37,16 @@ def get_or_create_user(session: Session, token_user: dict[str, str]) -> dict:
...
-    user = session.query(User).filter(User.email == email).first()
+    if external_id:
+        user = user_crud.get_user_by_external_id(session, external_id)
+    else:
+        user = session.query(User).filter(User.email == email).first()
 
     # Create User if DNE
     if not user:
-        db_user = User(fullname=fullname, email=email)
+        db_user = User(fullname=fullname, email=email, external_id=external_id)
```
raphael-cohere, line 49: We should probably not create users if scim is activated.

src/backend/services/context.py
raphael-cohere: okay managed to fix this nicely by checking if "bearer" is in header. We use "basic" for scim.

src/backend/routers/scim.py
```diff
@@ -0,0 +1,259 @@
...
+            "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
+        },
+    )
+
+
+@router.get("/Users")
+async def get_users(
+    session: DBSessionDep,
+    count: int = 100,
+    start_index: int = 1,
```
BeatrixCohere, line 53: Why 1 not 0?

src/backend/routers/scim.py
```diff
@@ -0,0 +1,259 @@
...
+    filter: Optional[str] = None,
+) -> ListUserResponse:
+    if filter:
+        single_filter = filter.split(" ")
+        filter_value = single_filter[2].strip('"')
```
BeatrixCohere, line 58: Should we do validation here?

src/backend/routers/scim.py
```diff
@@ -0,0 +1,259 @@
...
+        itemsPerPage=count,
+        Resources=groups,
+    )
+
+
+@router.get("/Groups/{group_id}")
+async def get_group(group_id: str, session: DBSessionDep):
+    db_group = group_crud.get_group(session, group_id)
+    if not db_group:
+        raise SCIMException(status_code=404, detail="Group not found")
```
BeatrixCohere, line 192: Can we log some of the exceptions?

src/backend/routers/scim.py
raphael-cohere: This is in the spec 🤷

src/backend/config/routers.py
```diff
@@ -131,4 +135,11 @@ class RouterName(StrEnum):
             Depends(validate_authorization),
         ],
     },
+    RouterName.SCIM: {
+        "default": [
+            Depends(get_session),
+            Depends(ScimAuthValidation()),
+        ],
+        "auth": [],
```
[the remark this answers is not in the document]
raphael-cohere: Auth dependencies only get add if user auth is setup, these endpoints should always be protected independently of that. Otherwise they would be open to the public if you do not have e.g.: OIDC configured.

src/backend/routers/scim.py
```diff
@@ -0,0 +1,259 @@
...
+        },
+    )
+
+
+@router.get("/Users")
+async def get_users(
+    session: DBSessionDep,
+    count: int = 100,
+    start_index: int = 1,
+    filter: Optional[str] = None,
```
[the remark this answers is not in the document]
raphael-cohere: For okta yes. That's all they support. Might be different for other providers.

src/backend/config/routers.py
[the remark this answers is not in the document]
raphael-cohere: misunderstand how the dependency system work, also added to auth now

src/backend/routers/scim.py
raphael-cohere: good idea, will do

src/backend/routers/scim.py
raphael-cohere: https://datatracker.ietf.org/doc/html/rfc7644
startIndex  The 1-based index of the first result in the current set
of list results.  REQUIRED when partial results are returned due
to pagination.

BeatrixCohere (approved):

PR #552, opened 2024-07-31
Add timestamp + descriptions to Alembic migrations

tianjing-li:
- Modifies base `make migration` command to accept a message param. eg `make migration message="Adding new table"` 
- Modifies `alembic.ini` template file to add YYYY_MM_DD_ prefix to all migrations
- Manually updates existing alembic migrations

PR #567, opened 2024-08-01
Agents: add private agents

lusmoura:

**AI Description**

This PR introduces the concept of private agents and the necessary changes to accommodate this feature. 

The `is_private` field is added to the `Agent` model, allowing agents to be marked as private. This field is also included in the `AgentBase` class, which is the base for the `Agent` model, and the `CreateAgentRequest` class, used for creating new agents. 

To support this change, the following modifications are made:
- The `get_agents` function is replaced with `get_public_agents` and `get_private_agents` functions in the `agent.py` file. These functions filter agents based on their privacy status, with `get_public_agents` excluding private agents and `get_private_agents` retrieving agents for a specific user.
- The `get_agent_by_name` function remains unchanged, but its description is updated to reflect that it only returns public agents.
- The `upgrade` and `downgrade` functions in `versions/998b82dfd78f_.py` are modified to add and remove the `is_private` column, respectively, from the "agents" table.
- The `get_public_and_private_agents` function is added to the `services/agent.py` file, combining public and private agents for a user.
- The `validate_user_has_access_to_agent` function is introduced to check if a user has access to a given agent, considering both public and private agents.
- The `list_agents`, `get_agent_by_id`, `update_agent`, `delete_agent`, and `list_agent_tool_metadata` functions in `routers/agent.py` are updated to incorporate the `validate_user_has_access_to_agent` function and handle private agents.
- The `list_tools` function in `routers/tool.py` is modified to include the `validate_user_has_access_to_agent` function.
- The `chat_stream` and `chat` functions in `routers/chat.py` are updated to validate user access to agents.
- The `search_conversations` function in `routers/conversation.py` is modified to validate user access to agents and handle private agents.
- The `process_chat` function in `services/chat.py` is updated to include the `validate_user_has_access_to_agent` function.
- The `get_files_by_agent_id` function in `services/file.py` is modified to validate agent existence and user access.
- The `validate_chat_request` function in `services/request_validators.py` is updated to use the `validate_agent_exists` and `validate_user_has_access_to_agent` functions.
- Test functions in `tests/crud/test_agent.py` and `tests/routers/test_agent.py` are adjusted to reflect the changes and test the new functionality.


src/backend/crud/agent.py
```diff
@@ -75,12 +75,42 @@ def get_agents(
     query = db.query(Agent)
 
     if organization_id is not None:
-        query = query.filter(Agent.organization_id == organization_id)
+        # Piblic agents -> is_private = False or None
```
BeatrixCohere, line 78:
```suggestion
        # Public agents -> is_private = False or None
```

src/backend/crud/agent.py
```diff
@@ -75,12 +75,42 @@ def get_agents(
...
+        query = query.filter(
+            Agent.organization_id == organization_id, not Agent.is_private
```
BeatrixCohere, line 80: Should we check for null?

src/backend/crud/agent.py
```diff
@@ -75,12 +75,42 @@ def get_agents(
...
+      offset (int): Offset of the results.
+      limit (int): Limit of the results.
+      organization_id (str): Organization ID.
+
+    Returns:
+      list[Agent]: List of agents.
+    """
+    query = (
+        db.query(Agent)
+        .filter(Agent.user_id == user_id, Agent.is_private is True)
```
BeatrixCohere, line 107: Same here for null

src/backend/routers/agent.py
```diff
@@ -336,13 +352,23 @@ async def list_agent_tool_metadata(
...
+        agent = agent_crud.get_agent_by_id(session, agent_id)
+        tool_metadata = (
+            agent_tool_metadata_crud.get_all_agent_tool_metadata_by_agent_id(
+                session, agent_id
+            )
         )
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e))
 
+    validate_user_has_access_to_agent(user_id, agent, agent_id=agent_id)
```
BeatrixCohere, line 369: Could we not do this in `get agent by id`?

src/backend/services/agent.py
```diff
@@ -42,3 +44,37 @@ def raise_db_error(e: Exception, type: str, name: str):
...
+
+
+def get_public_and_private_agents(
+    session: DBSessionDep,
+    user_id: str,
+    organization_id: Optional[str] = None,
+    offset: int = 0,
+    limit: int = 100,
+) -> list[Agent]:
+    public_agents = agent_crud.get_public_agents(
```
BeatrixCohere, line 56: Could we have one function with a filter for private?

src/backend/services/agent.py
```diff
@@ -42,3 +44,37 @@ def raise_db_error(e: Exception, type: str, name: str):
...
+    # Check if agent exists and user has access to it
+    error = None
+    if not agent:
+        error = f"Agent with ID: {agent_id} not found."
+
+    if agent and agent.user_id != user_id and agent.is_private:
+        error = f"Agent with ID: {agent.id} not found."
+
+    if error:
+        raise HTTPException(status_code=404, detail=error)
```
BeatrixCohere, line 78: Would this cause an exception when just getting all the agents?

src/backend/routers/agent.py
lusmoura: I was trying to keep the crud functions atomic, but yeah I can change the filter there

src/backend/services/agent.py
lusmoura: Do you mean a crud function?

src/backend/services/agent.py
lusmoura: I don't think it would 🤔

src/backend/database_models/agent.py
```diff
@@ -18,6 +18,7 @@ class Agent(Base):
 
     tools: Mapped[list[str]] = mapped_column(JSON, default=[], nullable=False)
     tools_metadata: Mapped[list[AgentToolMetadata]] = relationship("AgentToolMetadata")
+    is_private: Mapped[bool] = mapped_column(Boolean, default=False, nullable=True)
```
tianjing-li, line 21: No need to make it nullable=True since you set a default

src/backend/alembic/versions/998b82dfd78f_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
```
tianjing-li, line 1: Could you update the migration to be in line with this PRs changes? https://github.com/cohere-ai/cohere-toolkit/pull/552

- Eg add timestamp and description, so `2024_08_01_998..f_add_agent_is_private.py` 
- Add a description instead of `empty message`

src/backend/crud/agent.py
```diff
@@ -74,16 +84,26 @@ def get_agents(
     """
     query = db.query(Agent)
 
+    if visibility == AgentVisibility.PUBLIC:
+        # Public agents -> is_private = False or None
+        query = query.filter((Agent.is_private == False) | (Agent.is_private == None))
```
tianjing-li, line 89: With the default, we can avoid making this nullable now - and remove the Agent.is_private == None logic here

scott-cohere, line 89: can remove None check 

re: what TJ said about removing nullable as we set the default to false.

nit: can remove these comments, the code is pretty self explanatory.

src/backend/tests/routers/test_agent.py
```diff
@@ -971,6 +1077,62 @@ def test_update_agent_invalid_tool(
...
+
+    response = session_client.put(
+        f"/v1/agents/{agent.id}", json=request_json, headers={"User-Id": user.id}
+    )
+    assert response.status_code == 200
+    updated_agent = response.json()
+    assert updated_agent["name"] == "updated name"
+    assert updated_agent["is_private"] == False
+
+
```
tianjing-li, line 1135: Can we update a private Agent to public and vice versa? Might be good to add some test scenarios for those (eg, a private -> public agent would now be accessible through crud by any user)

tianjing-li (approved):
Couple small requests - otherwise ready to go

src/backend/tests/routers/test_agent.py
tianjing-li: If no bandwidth feel free to make a ticket a prioritize as needed, not necessary to get in but would be good coverage

src/backend/database_models/agent.py
lusmoura: Wouldn't it break for existing rows in the DB tho? Or do I need either a default or nullable?

src/backend/tests/routers/test_agent.py
lusmoura: Done!

PR #569, opened 2024-08-01
backend: upload + chat with compass

scott-cohere:
This PR introduces using Compass as a storage solution for files. Here are some of the key changes:

- In the file service, adding a ton of logic and functions to query Compass for our needs instead of the DB.
- Splitting uploads into two different categories (conversation vs. agent). 
- Files are indexed into Compass under a conversation ID or agent ID. This makes things a bit complicated when want to retrieve files as we also need the index. Made changes to accomodate this, left comments for clarity in PR.
- For files uploaded for an agent, they are indexed into temporary indices then when an agent is created they are consolidated into one index (agent ID).

**AI Description**

This PR introduces changes related to file handling and storage in the backend of an application. Here's a summary of the modifications:

   - Added `conversation_id` to the `call_tools` function.
- **src/backend/config/settings.py**:
   - Introduced a new field, `use_compass_file_storage`, in the `FeatureFlags` class.
- **src/backend/config/tools.py**:
   - Modified the `ToolName` class:
     - Changed the `filenames` field to `files`, now a list of tuples containing the filename and file ID.
     - Updated the `ReadFileTool` parameter definition to use the new `file` field instead of `filename`.
- **src/backend/routers/agent.py**:
   - Added imports for `RequestFile`, `UploadFile`, and `Settings` from `fastapi`.
   - Included imports for `ToolName` from `backend.config.tools` and `DeleteFileResponse`, `UploadFileResponse` from `backend.schemas.file`.
   - Introduced functions for batch file upload and file deletion:
     - `batch_upload_file`: Allows uploading multiple files and returns a list of `UploadFileResponse`.
     - `delete_file`: Deletes a file by its ID.
   - Added logic to consolidate agent files into a single index in Compass.
- **src/backend/routers/chat.py**:
   - Removed `file_paths` from the `chat_stream` and `chat` functions.
- **src/backend/routers/conversation.py**:
   - Added imports for `Settings` from `backend.config.settings` and `Compass` from `backend.services.compass`.
   - Modified the `delete_conversation` function to use the `validate_conversation` function and updated file deletion logic.
   - Updated the `upload_file` function to check if Compass file storage is enabled before validating file size.
   - Adjusted the `batch_upload_file` function to use the `ctx` parameter and updated the file creation process.
   - Removed the `update_file` function.
- **src/backend/schemas/file.py**:
   - Made `conversation_id` optional in the `File` class.
   - Introduced a new optional field, `file_content`, in the `File` class.
   - Removed the `get_file_service` import.
   - Removed `file_paths` from the `process_chat` and `create_message` functions.
   - Deleted the `handle_file_retrieval` function.
- **src/backend/services/compass.py**:
   - Modified the `_raw_parsing` function to include a new parameter, `custom_context`.
- **src/backend/services/file.py**:
   - Added imports for `uuid`, `datetime`, `Settings`, `Context`, `Compass`, and `get_context`.
   - Introduced the `get_file_service` and `get_compass` functions to initialize instances of `FileService` and `Compass`.
   - Modified the `FileService` class:
     - Added a property, `is_compass_enabled`, to check if Compass is enabled for file storage.
     - Updated the `create_conversation_files` and introduced the `create_agent_files` function to support Compass file storage.
     - Adjusted the `get_files_by_agent_id` function to import `ToolName` from `backend.config.tools`.
     - Modified the `get_files_by_conversation_id` function to support Compass file storage.
     - Introduced the `delete_conversation_file_by_id` and `delete_agent_file_by_id` functions to handle file deletion.
     - Removed the `get_file_by_id`, `get_files_by_ids`, `update_file`, and `bulk_delete_files` functions.
     - Added new functions for Compass operations: `delete_file_in_compass`, `get_files_in_compass`, `consolidate_agent_files_in_compass`, and `insert_files_in_compass`.
   - Adjusted the `validate_file` function to support Compass file storage.
   - Introduced the `insert_files_in_db` function to insert files into the database.
   - Added the `attach_conversation_id_to_files` function to attach a conversation ID to files.
- **src/backend/tests/conftest.py**:
   - Added a new fixture, `mock_compass_settings`, to mock Compass-related settings.
- **src/backend/tests/routers/test_conversation.py**:
   - Modified the `test_list_files` function to use the `mock_compass_settings` fixture and updated file handling.
   - Adjusted the `test_upload_file_existing_conversation` function to use `mock_compass_settings` and updated file retrieval logic.
   - Updated the `test_batch_upload_file_existing_conversation` function to use `mock_compass_settings` and modified file handling.
   - Modified the `test_batch_upload_file_nonexistent_conversation_creates_new_conversation` function to use `mock_compass_settings` and updated file handling.
   - Adjusted the `test_delete_file` function to use `mock_compass_settings` and modified file retrieval.
- **src/backend/tools/files.py**:
   - Added imports for `SearchFilter`, `Settings`, `Compass`, `get_compass`, and `get_file_service`.
   - Introduced the `compass_file_search` function to search for files in Compass.
   - Modified the `ReadFileTool` and `SearchFileTool` classes to support Compass file storage and introduced a `SEARCH_LIMIT` attribute.


src/backend/chat/custom/custom.py
```diff
@@ -302,7 +303,7 @@ def add_files_to_chat_history(
             num_words = min(25, word_count)
             preview = " ".join(file.file_content.split()[:num_words])
 
-            files_message += f"Filename: {file.file_name}\nWord Count: {word_count} Preview: {preview}\n\n"
+            files_message += f"Filename: {file.file_name}\nFile ID: {file.id}\nWord Count: {word_count} Preview: {preview}\n\n"
```
scott-cohere, line 306: Files are indexed into Compass with either convo or agent ID, we can't retrieve a file just with the file name. File search and read tools will take in a tuple (file name, file id) instead.

src/backend/config/tools.py
```diff
@@ -82,9 +82,9 @@ class ToolName(StrEnum):
         display_name="Read Document",
         implementation=ReadFileTool,
         parameter_definitions={
-            "filename": {
-                "description": "The name of the attached file to read.",
-                "type": "str",
+            "file": {
+                "description": "A file represented as a tuple (filename, file ID) to read over",
+                "type": "tuple[str, str]",
```
scott-cohere, line 87: https://github.com/cohere-ai/cohere-toolkit/pull/569/files#r1711452685

src/backend/routers/agent.py
```diff
@@ -98,6 +109,30 @@ async def create_agent(
                 await update_or_create_tool_metadata(
                     created_agent, tool_metadata, session, ctx
                 )
+
+        # Consolidate agent files into one index in compass
```
scott-cohere, line 113: Upload and agent creation are two steps. 

Upload indexes agent files into individual orphaned indices. When an agent is being created with those file IDs as the tools metadata we consolidate them under one index (agent ID) here.

src/backend/routers/agent.py
```diff
@@ -476,6 +511,62 @@ async def delete_agent_tool_metadata(
     return DeleteAgentToolMetadata()
 
 
+@router.post("/batch_upload_files", response_model=list[UploadFileResponse])
```
scott-cohere, line 514: I decided to separate upload into two endpoints.

One used in a conversation, this one for the agent creation screen.

Rationale for this is that depending which kind of upload we do there are side effects. For conversation file upload we need to create conversations if they don't exists. Here we don't need to do that.

Also the semantics of `/conversation/upload` doesn't really make sense if we are uploading for an agent.

src/backend/routers/agent.py
```diff
@@ -476,6 +511,62 @@ async def delete_agent_tool_metadata(
...
+        )
+    except Exception as e:
+        raise HTTPException(
+            status_code=500, detail=f"Error while uploading agent file(s): {e}."
+        )
+
+    return uploaded_files
+
+
+@router.delete("/{agent_id}/files/{file_id}")
```
scott-cohere, line 539: Same here, deleting the _agents_ file, not a conversation file.

src/backend/routers/chat.py
```diff
@@ -165,7 +162,6 @@ async def chat(
         CustomChat().chat(
             chat_request,
             stream=False,
-            file_paths=file_paths,
```
scott-cohere, line 168: I'll follow up with a PR to remove file path everywhere. Tracking here https://linear.app/cohereai/issue/TLK-967/remove-file-paths-from-all-be-code

src/backend/routers/conversation.py
```diff
@@ -351,7 +352,9 @@ async def upload_file(
     """
 
     user_id = ctx.get_user_id()
-    validate_file_size(session, user_id, file)
+    # Currently do not limit file size for Compass
```
scott-cohere, line 355: @GangGreenTemperTatum Pointing this out here.

There isn't really a great way to check do this validation for Compass storage. We decided to just skip this validation for the moment. (GDrive files also does not have this check). Also since files can be uploaded to agents which are shared, we need to think about how that contributes to a user's limit...

Will probably have to do something where we track and store file sizes in the DB in `user` table manually or something

src/backend/routers/conversation.py
```diff
@@ -496,42 +504,6 @@ async def list_files(
     return files_with_conversation_id
 
 
-@router.put("/{conversation_id}/files/{file_id}", response_model=FilePublic)
-async def update_file(
```
scott-cohere, line 500: I don't think updating files is used at all. To reduce code and overhead of testing it I've just removed it.

@knajjars to confirm?

src/backend/schemas/file.py
```diff
@@ -10,7 +10,8 @@ class File(BaseModel):
     updated_at: datetime.datetime
 
     user_id: str
-    conversation_id: str
+    conversation_id: Optional[str] = None
+    file_content: Optional[str] = None
```
scott-cohere, line 14: This schema is used for both the request but also the type to return for methods in the file service (since if we use Compass we can't map it to the DB file model schema).

src/backend/services/chat.py
```diff
@@ -310,29 +308,6 @@ def create_message(
     return message
 
 
-def handle_file_retrieval(
```
scott-cohere, line 313: Again, legacy due to use of file paths before.

src/backend/tests/conftest.py
```diff
@@ -204,3 +205,19 @@ def mock_available_model_deployments(request):
...
+        mock_settings.feature_flags.use_compass_file_storage = os.getenv(
+            "ENABLE_COMPASS_FILE_STORAGE", False
+        )
+        mock_settings.tools.compass.api_url = os.getenv("COHERE_COMPASS_API_URL")
+        mock_settings.tools.compass.api_parser_url = os.getenv(
+            "COHERE_COMPASS_API_PARSER_URL"
+        )
+        mock_settings.tools.compass.username = os.getenv("COHERE_COMPASS_USERNAME")
+        mock_settings.tools.compass.password = os.getenv("COHERE_COMPASS_PASSWORD")
+        yield mock_settings
```
scott-cohere, line 223: Seems like when `Settings()` is called in the code during a pytest there may be some issues with its instantiation.

This overrides the results when it is called for compass settings. Also might be useful if we ever want to run the compass tests with on a test instance with different credentials, etc.

Though, after fixing and running tests, they are very slow. We've faced this issue with testing on the test DB. Perhaps time to start investing into mocking instead of using these direct calls on storage. For now this will suffice.

src/backend/chat/custom/custom.py
tianjing-li: @GangGreenTemperTatum yes

docker-compose.yml
```diff
@@ -110,14 +110,14 @@ services:
...
-    networks:
-      - proxynet
+  # terrarium:
+  #   image: ghcr.io/cohere-ai/terrarium:latest
+  #   ports:
+  #     - "8080:8080"
+  #   expose:
+  #     - "8080"
+  #   networks:
+  #     - proxynet
```
tianjing-li, line 120: Should this be reverted?

src/backend/config/secrets.template.yaml
```diff
@@ -36,3 +36,8 @@ auth:
     client_id:
     client_secret:
     well_known_endpoint:
+compass:
```
tianjing-li, line 39: Should these new configs be added to the `settings.py` as well

src/backend/routers/agent.py
```diff
@@ -101,6 +112,31 @@ async def create_agent(
...
+        # Consolidate agent files into one index in compass
+        if (
+            Settings().feature_flags.use_compass_file_storage
+            and created_agent.tools_metadata
+        ):
+            artifacts = next(
+                (
+                    tool_metadata.artifacts
+                    for tool_metadata in created_agent.tools_metadata
+                    if tool_metadata.tool_name == ToolName.Read_File
```
tianjing-li, line 125: Suggestion: Use an array and compare `if .. in`

src/backend/routers/agent.py
```diff
@@ -101,6 +112,31 @@ async def create_agent(
...
+                    or tool_metadata.tool_name == ToolName.Search_File
+                ),
+                [],
+            )
+            file_ids = list(
+                set(
+                    artifact.get("id")
+                    for artifact in artifacts
+                    if artifact.get("type") == "local_file"
```
tianjing-li, line 134: Are the artifact `type` values stored in some enum somewhere? I think ideally we compare to an enum. Not too familiar with what the possible values can be

src/backend/routers/agent.py
```diff
@@ -101,6 +112,31 @@ async def create_agent(
...
+                )
+            )
+            if len(file_ids) > 0:
```
tianjing-li, line 137: suggestion: `if file_ids:` will capture this

src/backend/routers/agent.py
```diff
@@ -615,6 +651,62 @@ async def delete_agent_tool_metadata(
...
+    except Exception as e:
+        raise HTTPException(
+            status_code=500, detail=f"Error while uploading agent file(s): {e}."
+        )
+
+    return uploaded_files
+
+
+@router.delete("/{agent_id}/files/{file_id}")
+async def delete_file(
```
tianjing-li, line 680: Suggestion: update method to `delete_agent_file` to be explicit

src/backend/routers/agent.py
```diff
@@ -615,6 +651,62 @@ async def delete_agent_tool_metadata(
...
+    agent_id: str,
+    file_id: str,
+    session: DBSessionDep,
+    ctx: Context = Depends(get_context),
+) -> DeleteFileResponse:
+    """
+    Delete a file by ID.
```
tianjing-li, line 687: Delete an Agent's file by file ID

src/backend/routers/chat.py
tianjing-li: #652 now does this 🥇

src/backend/config/secrets.template.yaml
tianjing-li: Nevermind! I see they're under tools if I'm not mistaken

tianjing-li (changes requested):
Some suggestions

src/backend/routers/conversation.py
scott-cohere: https://linear.app/cohereai/issue/TLK-987/investigate-limiting-file-sizes-for-uploads

docker-compose.yml
```diff
@@ -110,14 +110,14 @@ services:
...
 
-  terrarium:
-    image: ghcr.io/cohere-ai/terrarium:latest
-    ports:
-      - "8080:8080"
-    expose:
-      - "8080"
-    networks:
-      - proxynet
+  # terrarium:
```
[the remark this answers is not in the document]
scott-cohere: Woops sorry, just removed it for local testing. Will revert when merging.

src/backend/routers/agent.py
```diff
@@ -98,6 +109,30 @@ async def create_agent(
...
+
+        # Consolidate agent files into one index in compass
+        if (
+            Settings().feature_flags.use_compass_file_storage
+            and created_agent.tools_metadata
+        ):
+            artifacts = next(
+                (
+                    tool_metadata.artifacts
+                    for tool_metadata in created_agent.tools_metadata
```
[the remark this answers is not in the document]
scott-cohere: @GangGreenTemperTatum not really too sure what you mean here by this but here. could you please elaborate bit? thanks

docker-compose.yml
scott-cohere: yep

src/backend/routers/agent.py
scott-cohere: they're meant to be flexible for whatever tool. that being said i think a good idea is to just add it an enum class for artifacts in each tools file. ill go and do that. 👍

src/backend/routers/agent.py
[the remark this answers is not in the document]
scott-cohere: i created a new file uploads endpoint for agents https://github.com/cohere-ai/cohere-toolkit/pull/569/files/f1a500b6bf3bcf199f855102d43f1d1b45d5d598#diff-aa70c2598055d0e9e93e707c540a2b7e4cbd7d0ac8baaaa84e0265b0fc0edbd3R514

the implementation is basically the same but we have extra logic for conversation file uploads (namely, creates a new convo if youre uploading on a fresh chat where the convo hasnt been created yet)

also for semantics doesn't make sense to have `/conversations/..upload` if its uploading for an agent.

tianjing-li (approved):
Looks good to me - good to merge after/if you've tested this logic

PR #573, opened 2024-08-02
fix(toolkit): API host name for RSC

knajjars:
React Server Components runs in the server, for our case this means that it runs in the same network that our backend is running, therefore to communicate with the backend we need to specify the internal docker network route to the backend (http://backend/). For the client bundle we still use the exposed port 8000 because requests are performed from an external network from that of docker.
- API_HOSTNAME -> to be used in RSC context
- NEXT_PUBLIC_API_HOSTNAME -> for the client bundle

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       - PGUSER=postgres
       - POSTGRES_PASSWORD=postgres
     ports:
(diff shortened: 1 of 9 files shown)


EugeneLightsOn (approved):
Looks good to me

tomtobac (approved):

PR #592, opened 2024-08-06
Add calling tools async with a timeout per tool call 

BeatrixCohere:
Add calling tools async with a timeout per tool call 

**AI Description**

This PR introduces a new module, `tool_calls.py`, in the `src/backend/chat/custom` directory. This module provides a new function, `async_call_tools`, which replaces the `call_tools` function in `custom.py. The `async_call_tools` function is now used by the `call_chat` function in `custom.py` to handle tool calls in the chat history.

The `async_call_tools` function:
- Accepts similar parameters to the previous `call_tools` function, including `chat_history`, `deployment_model`, `ctx`, and `**kwargs`.
- Checks for tool calls in the `chat_history` and returns an empty list if none are found.
- Uses the `_call_all_tools_async` function to call all tools asynchronously and gather their results.
- Handles potential timeouts during tool calls, raising an `HTTPException` with a status code of 500 and a detailed error message if a timeout occurs.
- Returns the tool results after ranking and chunking them.

The `call_chat` function in `custom.py` has been updated to use the new `async_call_tools` function, replacing the previous call to `self.call_tools`.

Additionally, new test cases have been added in `test_tool_calls.py` to verify the functionality of the `async_call_tools` function, including:
- `test_async_call_tools_success`: Tests a successful tool call and asserts the expected results.
- `test_async_call_tools_failure`: Tests a failing tool call and asserts the expected HTTPException with a status code of 500.
- `test_async_call_tools_timeout`: Tests a tool call that exceeds the timeout and asserts the expected HTTPException.
- `test_async_call_tools_failure_and_success`: Tests a mix of failing and successful tool calls and asserts the appropriate error handling and results.

Overall, this PR introduces asynchronous tool handling and improves error handling, enhancing the chat functionality's robustness and performance.


lusmoura (commented):
I remember a conversation about the need to make sure the tool results are listed in the same order as the tool calls or something. Do you know if it's still true? And if so does this PR ensure the results are ordered?

lusmoura (approved):

PR #596, opened 2024-08-06
Restore chat flow

EugeneLightsOn:
- Restore chat flow

**AI Description**

This PR makes changes to the backend of the application, including database models, chat processing, and testing. 

## Summary
- Updates the `src/backend/database_models/seeders/deplyments_models_seed.py` file to use SQL commands for inserting data into the `deployments` and `models` tables, replacing the previous method of using the `Session` object. 
- Removes some code related to `default_user` and `default_organization` in the `deployments_models_seed` function. 
- Removes the `default_agent` code block and associated code from the same function. 
- Introduces the `uuid4` function from the `uuid` module to generate unique IDs for the `deployment_id` and `model_id`. 
- Updates the `src/backend/services/chat.py` file by removing some code related to deployment configuration and agent association. 
- Removes a block of code that sets a temporary note about not setting the model and instead uses the default model. 
- Updates the `src/backend/tests/routers/test_agent.py` file by changing the expected number of response agents from 2 to 1 in the `test_list_agents_with_pagination` test. 
- Updates the `src/backend/tests/routers/test_chat.py` file by adding new test cases and modifying existing ones to include the creation of `User`, `Agent`, `Deployment`, `Model`, and `AgentDeploymentModel` objects. 

## Details:
- **src/backend/database_models/seeders/deplyments_models_seed.py**:
  - The `uuid` module is imported to generate unique IDs.
  - The `sqlalchemy.text` module is imported for executing SQL commands.
  - The `Deployment`, `Model`, and `Organization` classes are imported from `backend.database_models`.
  - The `Agent`, `AgentDeploymentModel`, `User`, and `ModelDeploymentName` classes are no longer imported.
  - The `deployments_models_seed` function now uses SQL commands to insert data into the `deployments` and `models` tables, with unique IDs generated using `uuid4`.
  - The code related to `default_user` and `default_organization` has been removed, along with the associated `default_organization.users.append(default_user)` line.
  - The code block for `default_agent` and associated code have been removed.

  - Removes code related to checking and setting deployment configuration and agent association.
  - Deletes a block of code that sets a temporary note about not setting the model and instead uses the default model.

- **src/backend/tests/routers/test_agent.py**:
  - Updates the expected number of response agents from 2 to 1 in the `test_list_agents_with_pagination` test.

- **src/backend/tests/routers/test_chat.py**:
  - Adds or modifies test cases to include the creation of `User`, `Agent`, `Deployment`, `Model`, and `AgentDeploymentModel` objects.
  - Updates the `test_streaming_new_chat_metrics_with_agent` test case to use the `get_factory` function to create the necessary objects.
  - Similar updates are made to the `test_streaming_new_chat_with_agent` and `test_streaming_new_chat_with_agent_existing_conversation` test cases.
  - The `test_streaming_chat_with_tools_not_in_agent_tools` test case is modified to use the `get_factory` function to create the required objects.

### DIFF
--- a/src/backend/database_models/seeders/deplyments_models_seed.py
+++ b/src/backend/database_models/seeders/deplyments_models_seed.py
@@ -1,22 +1,19 @@
 import os
+import json
+from uuid import uuid4
 
 from dotenv import load_dotenv
+from sqlalchemy import text
 from sqlalchemy.orm import Session
 
 from backend.config.deployments import ALL_MODEL_DEPLOYMENTS, ModelDeploymentName
 from backend.database_models import (
-    Agent,
-    AgentDeploymentModel,
     Deployment,
     Model,
     Organization,
-    User,
 )
 from community.config.deployments import (
     AVAILABLE_MODEL_DEPLOYMENTS as COMMUNITY_DEPLOYMENTS_SETUP,
-)
-from community.config.deployments import (
-    ModelDeploymentName as CommunityModelDeploymentsName,
 )
 
 load_dotenv()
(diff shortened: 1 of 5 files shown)


src/backend/database_models/seeders/deplyments_models_seed.py
```diff
@@ -112,86 +105,63 @@ def deployments_models_seed(op):
...
-            "clinical_trials",
-        ],
-        user_id=default_user.id,
-        organization_id=default_organization.id,
-    )
-    session.add(default_agent)
-    session.commit()
 
     # Seed deployments and models
     for deployment in MODELS_NAME_MAPPING.keys():
```
BeatrixCohere, line 116: I don't think we want to seed with any data?

scott-cohere: hmm why not?

removing the default agent w/ user-id fixes the migration issues, and also removed the default agent as discussed. we should still seed the model configs, no?

scott-cohere (approved):

src/backend/database_models/seeders/deplyments_models_seed.py
EugeneLightsOn: I believe we need to keep the models and deployments in the database. I removed the default agent as we discussed, but we still need the models and deployments. I also refactored the seeder to use raw SQL to avoid conflicts with SQLAlchemy model changes.

giannis2two (approved):

PR #608, opened 2024-08-07
Agents: add private agents

lusmoura:
Closed #567 because it would be easier to start from scratch than to try to incorporate recent changes.

This PR does NOT cover updating the visibility of an agent, since this will require some more discussion.

**AI Description**

This PR introduces a new is_private field to the Agent model, allowing agents to be marked as either private or public. This change affects the CRUD operations for agents, as well as the agent-related routers and services.

## Summary of Changes
- Added an is_private column to the agents table.
- Updated the Agent model to include an is_private field.
- Modified agent CRUD operations to handle the new is_private field.
- Added AgentVisibility enum to represent agent visibility levels (private, public, all).
- Updated agent-related routers and services to consider agent visibility.
- Adjusted tests to account for the new is_private field and visibility levels.


src/backend/schemas/agent.py
```diff
@@ -7,6 +8,12 @@
 from backend.schemas.deployment import ModelSimple as ModelSchema
 
 
+class AgentVisibility(str, Enum):
+    PRIVATE = "private"
+    PUBLIC = "public"
+    ALL = "all"
```
tomtobac, line 14: isn't the same, public / all?

lusmoura: `All` is supposed to be `public + private`. So if a user has private agents, you could get all the agents they have access to by using `All`.

src/backend/alembic/versions/2024_08_07_71d7b35c8a44_.py
```diff
@@ -0,0 +1,43 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "agents",
+        sa.Column(
+            "is_private", sa.Boolean(), nullable=False, server_default=sa.false()
```
BeatrixCohere, line 27: Confirming this converts existing to have false?

src/backend/crud/agent.py
```diff
@@ -52,9 +62,15 @@ def get_agent_by_name(db: Session, agent_name: str) -> Agent:
     Returns:
       Agent: Agent with the given name.
     """
-    return db.query(Agent).filter(Agent.name == agent_name).first()
+    agent = db.query(Agent).filter(Agent.name == agent_name).first()
 
+    if agent and agent.is_private and agent.user_id != user_id:
```
BeatrixCohere, line 67: @raphael-cohere will this work with your changes?

src/backend/schemas/agent.py
```diff
@@ -7,6 +8,12 @@
 from backend.schemas.deployment import ModelSimple as ModelSchema
 
 
+class AgentVisibility(str, Enum):
```
BeatrixCohere, line 11: nit: StrEnum

BeatrixCohere (approved):

src/backend/alembic/versions/2024_08_07_71d7b35c8a44_.py
lusmoura: Yep! I tested it locally and it worked.

src/backend/schemas/agent.py
```diff
@@ -93,6 +100,7 @@ class CreateAgentRequest(BaseModel):
     # deployment_id or deployment_name
     deployment: str
     organization_id: Optional[str] = None
+    is_private: Optional[bool] = False
```
EugeneLightsOn, line 104: Do we need to add this to the update schema?

EugeneLightsOn (commented):
Looks good to me,  one thing - please take a look at the agent update schema.

src/backend/schemas/agent.py
lusmoura: Will add updating agents in a separate PR!

EugeneLightsOn: Thanks

EugeneLightsOn (approved):
Looks good to me.

PR #610, opened 2024-08-08
feat(toolkit): sort agents in groups

knajjars:
Sorts agents in groups

https://github.com/user-attachments/assets/1d37a43f-fcb7-4f05-a8c2-961c01ffad1b

**AI Description**

This PR introduces a new `user_id` field to the `$AgentPublic` and `AgentPublic` types, and updates the `DiscoverAgentCard` and `DiscoverAgents` components to use this new field. 

- The `user_id` field is added to the `$AgentPublic` type in `schemas.gen.ts`.
- The `user_id` field is also added to the `AgentPublic` type in `types.gen.ts`.
- The `DiscoverAgentCard` component now uses the `AgentPublic` type for its `agent` prop, and includes a `user_id` field.
- The `DiscoverAgents` component now uses the `AgentPublic` type for its `agents` prop, and includes a `user_id` field.
- The `DiscoverAgents` component also imports the `useSession` hook and filters the `agents` array to include only those created by the current user.
- The `ConversationList` and `ConversationListLoading` components now provide a default empty array for the `conversations` data.
- A new 'lock' icon is added to the `IconList` in `Icon.tsx`.


src/interfaces/assistants_web/src/components/Agents/DiscoverAgents.tsx
```diff
@@ -1,28 +1,20 @@
...
-    <Text>Private</Text>
-  </div>,
-];
+const GROUPED_ASSISTANTS_LIMIT = 15;
 
-const BASE_AGENTS: Array<AgentPublic & { isBaseAgent: boolean }> = [
+const BASE_AGENTS: Array<AgentPublic> = [
   {
-    id: '',
+    id: 'default',
```
tomtobac, line 17: if you add the `default` as an id it will have another bg color than the agent list

src/interfaces/assistants_web/src/components/Agents/DiscoverAgentCard.tsx
```diff
@@ -1,68 +1,98 @@
...
-              <Text className={cn('uppercase text-white', contrastText)} styleAs="p-lg">
-                {name[0]}
+              <Text className={cn('uppercase', contrastText)} styleAs="p-lg">
+                {agent?.name[0]}
               </Text>
             )}
+            {isPrivate && (
+              <span className="absolute bottom-0 right-0 translate-x-1/2 translate-y-1/2 transform">
+                <Icon
+                  name="lock"
```
tomtobac, line 62: James agreed that is may not be necessary

src/interfaces/assistants_web/src/components/Agents/DiscoverAgents.tsx
knajjars: no it should not

tomtobac (approved):

PR #612, opened 2024-08-08
refactor: hotkeys supports group name

tomtobac:
## Description

Refactor HotKeys command bar to align with the future designs
- Supports group name
- Split QuickActions in different groups

### Screenshots

image

image


src/interfaces/assistants_web/src/components/Shared/HotKeys/HotKeysProvider.tsx
```diff
@@ -45,26 +35,23 @@ export const HotKeysProvider: React.FC<HotKeysProviderProps> = ({ customHotKeys
...
-    commands,
-    action,
-  }));
-
-  return <HotKeysDialog isOpen={isDialogOpen} close={close} customActions={dialogCustomActions} />;
+  hotKeys
+    .map((hk) => hk.quickActions)
+    .flat()
+    .forEach(({ commands, action, options, dependencies }) => {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
knajjars, line 42: not for now, but i think we should check a better way to register hooks outside of a forloop

knajjars (approved):

PR #629, opened 2024-08-09
CI: fix check duplicate revision id

lusmoura:
The migrations files have a date before the revision id now, so splitting the filenames is breaking the check.

**AI Description**

The `main` function in `check_revision_ids.py` has been modified to check for duplicate file names instead of revision IDs.

- The `revision_ids` list is no longer created by extracting the first part of each file name before the underscore.
- The condition to raise an exception has been changed to check if the length of `files` is not equal to the length of a set of `files`, indicating duplicate file names.
- The exception message has been updated to reflect the change in logic, now raising an exception if a "Duplicate file name" is found.
- The print statement remains the same, indicating that "All revision IDs are unique."

### DIFF
--- a/.github/scripts/check_revision_ids.py
+++ b/.github/scripts/check_revision_ids.py
@@ -3,10 +3,9 @@ import glob
 
 def main():
     files = glob.glob("src/backend/alembic/versions/*.py")
-    revision_ids = [file.split("_")[0] for file in files]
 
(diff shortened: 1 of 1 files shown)


EugeneLightsOn (approved):
Looks good to me

BeatrixCohere (approved):

PR #643, opened 2024-08-12
fix(toolkit): only show base tools for base assistant

knajjars:
return BASE_AGENT when using the `useAgent()` hook and no ID is passed.

**AI Description**

This pull request makes changes to the assistant web interface, specifically to the `DataSourceMenu`, `ToolEvents`, and `Welcome` components, as well as updates to the `BASE_AGENT` constant and the `useAgent` hook. 

- In the `DataSourceMenu.tsx` file, the `className` prop for the `div` element now includes dynamic styling based on the presence of a tool with a matching name in the `paramsTools` array. The `agent` prop is also checked for an `id` property before being used.
- The `ToolEvents.tsx` file has a small change to the `className` prop of the outer `div` element, where the value `dark:bg-volcanic-200` has been updated to `dark:bg-volcanic-200`.
- The `Welcome.tsx` component now checks for a base agent by verifying if the `id` property of the `agent` prop is an empty string. This check is used to conditionally render the component's content, including the agent's name, description, and "By Cohere" caption.
- The `BASE_AGENT` constant in `constants.ts` has been updated to include a new description, "Ask questions and get answers based on your files," and the `tools` array now includes the values from `AGENT_SETTINGS_TOOLS` and `FILE_UPLOAD_TOOLS`.
- The `useAgent` hook in `agents.ts` now returns the `BASE_AGENT` constant if the `agentId` is undefined, instead of throwing an error.


src/interfaces/assistants_web/src/components/Welcome.tsx
```diff
@@ -23,7 +23,7 @@ export const Welcome: React.FC<Props> = ({ show, agentId }) => {
   const { data: tools = [], isLoading: isToolsLoading } = useListTools();
   const { contrastText, bg, contrastFill } = useBrandedColors(agentId);
 
-  const isAgent = agentId !== undefined && !isAgentsLoading && !!agent;
+  const isBaseAgent = agent?.id === '';
```
tomtobac, line 26: we should create a utility function for this

tomtobac (approved):

src/interfaces/assistants_web/src/components/Welcome.tsx
knajjars: yeah will do

PR #648, opened 2024-08-12
backend: local unit tests

mkozakov:
**Description**

- Add `make install` target for installing python dependencies
- Add `make test-db` and ` target for spinning up a postgres DB in docker to be used in unit tests
- Update pytest config to point to test-db
- Update `make run-unit-tests` target to run `pytest` locally
- Update github action to run tests locally instead of in a docker container
- Move redis and compass tests from Unit to Integration


elliottsj (approved):
🐸 looks great.

I would also suggest updating the README so that devs know how to run the tests.

i.e.
- how to invoke run-unit-tests to run all tests, versus individual tests
- when to run test-db, migrate-test

src/backend/alembic/versions/2024_07_16_ed17f144f4bf_add_default_user.py
```diff
@@ -11,7 +11,8 @@
 import sqlalchemy as sa
 from alembic import op
 
-from src.backend.schemas.user import DEFAULT_USER_ID, DEFAULT_USER_NAME
+DEFAULT_USER_ID = "user-id"
```
lusmoura, line 14: Just out of curiosity, why are we redefining these variables here?

lusmoura (approved):
🚀

src/backend/alembic/versions/2024_07_16_ed17f144f4bf_add_default_user.py
mkozakov: migrations should never import from the main code, since the main code can change, where-as migrations are like constants.

PR #654, opened 2024-08-13
google drive tool: new endpoint for google drive tool auth deletion

chantelle-cohere:
### Description
This PR adds a new endpoint which handles deleting google drive tool auth.
HTTP Request Method: `DELETE`
Path: `/v1/tool/auth`
Request Body:
```json
{
    "tool_id": "<tool_id_to_be_deleted>"
}
```
Note that currently, the only supported `tool_id` is `google_drive`.

Expected Response:
```json
{
    "status": "success"
}
```

### Issue
fix TLK-875

### Dependencies
None

### TODO

**AI Description**

This PR introduces a new endpoint for deleting tool authentication and updates the `GoogleDriveAuth` class to include a `delete_tool_auth` method.

## Summary
The changes in this PR allow for the deletion of tool authentication, specifically for Google Drive. The new endpoint, `/tool/auth`, is added to the router in `src/backend/routers/auth.py`. It defines the logic for removing the corresponding ToolAuth from the database for the requesting user. Additionally, the `GoogleDriveAuth` class in `src/backend/tools/google_drive/auth.py` is updated to include a `delete_tool_auth` method, which handles the deletion process.

## Changes
- Introduce a new endpoint, `/tool/auth`, for deleting tool authentication.
- Implement the `delete_tool_auth` method in the `GoogleDriveAuth` class.

### DIFF
--- a/src/backend/routers/auth.py
+++ b/src/backend/routers/auth.py
@@ -2,14 +2,14 @@ import json
 from typing import Union
 from urllib.parse import quote
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi.responses import RedirectResponse
 from starlette.requests import Request
 
 from backend.config.auth import ENABLED_AUTH_STRATEGY_MAPPING
 from backend.config.routers import RouterName
 from backend.config.settings import Settings
-from backend.config.tools import AVAILABLE_TOOLS
+from backend.config.tools import AVAILABLE_TOOLS, ToolName
 from backend.crud import blacklist as blacklist_crud
 from backend.database_models import Blacklist
 from backend.database_models.database import DBSessionDep
@@ -317,3 +317,69 @@ async def login(
     response = RedirectResponse(redirect_uri)
 
     return response
+
+
+@router.delete("/tool/auth")
+async def delete_tool_auth(
+    request: Request, session: DBSessionDep, ctx: Context = Depends(get_context)
+):
+    """
+    Endpoint to delete Tool Authentication.
+
+    If completed, the corresponding ToolAuth for the requesting user is removed from the DB.
+
+    Args:
+        request (Request): current Request object. Note that the request body should contain the tool_id parameter as the tool to delete.
+        session (DBSessionDep): Database session.
+        ctx (Context): Context object.
+
+    Returns:
+        {"status": "success"} on successful deletion.
+
+    Raises:
+        HTTPException: If there was an error deleting the tool auth.
+    """
+
+    logger = ctx.get_logger()
+    logger.error(event=f"CHANTELLE TEST request body {request.body}")
+
+    def log_and_return_error(error_message: str):
+        logger.error(event=error_message)
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"{error_message}",
+        )
+
+    user_id = ctx.get_user_id()
+    request_body = await request.json()
+    tool_id = request_body["tool_id"]
+
+    logger.error(event=f"CHANTELLE TEST tool_id {tool_id}, user_id {user_id}")
+
+    if user_id is None or user_id == "" or user_id == "default":
+        log_and_return_error("User ID not found.")
+    
+    if not isinstance(tool_id, str):
+        log_and_return_error("tool_id must be present in the request body and must be a string.")
+
+    if tool_id != ToolName.Google_Drive:
+        log_and_return_error(f"Deletion for {tool_id} not implemented.")
+
+    tool = AVAILABLE_TOOLS.get(tool_id)
+
+    if tool.auth_implementation is None:
+        log_and_return_error(f"Tool {tool.name} does not have an auth_implementation required for Tool Auth Deletion.")
+
+    try:
+        tool_auth_service = tool.auth_implementation()
+        tool_auth_service.delete_tool_auth(session, user_id)
+        is_delete_tool_auth_successful = tool_auth_service.delete_tool_auth(session, user_id)
+
+        if not is_delete_tool_auth_successful:
+            log_and_return_error("Error deleting Tool Auth.")
+    
+    except Exception as e:
+        log_and_return_error(str(e))
+
+
+    return {"status": "success"}
--- a/src/backend/routers/auth.py
+++ b/src/backend/routers/auth.py
@@ -341,7 +341,6 @@ async def delete_tool_auth(
     """
 
     logger = ctx.get_logger()
-    logger.error(event=f"CHANTELLE TEST request body {request.body}")
 
     def log_and_return_error(error_message: str):
         logger.error(event=error_message)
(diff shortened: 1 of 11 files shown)


chantelle-cohere:
The new endpoint currently returns `{"status": "success"}` on successful tool auth deletion. Should it return anything else?
CC @tomtobac / @knajjars / @BeatrixCohere

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+        logger.error(event=error_message)
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"{error_message}",
+        )
+
+    user_id = ctx.get_user_id()
+
+    try:
+        request_body = await request.json()
```
tomtobac, line 355: hey @chantelle-cohere, a delete shouldn't have a body request. 

> DELETE cannot have request body, just like GET cannot have a request body. (Well, it's a syntactically valid HTTP message, but there's no semantic meaning to the body.)

I think we should pass the Id of the tool as a parameter and keep the delete method.

e.g.:

```
DELETE /tool/auth/<uuid>
```

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+
+    if user_id is None or user_id == "" or user_id == "default":
+        log_and_return_error("User ID not found.")
+
+    if not isinstance(tool_id, str):
+        log_and_return_error(
+            "tool_id must be present in the request body and must be a string."
+        )
+
+    if tool_id != ToolName.Google_Drive:
```
BeatrixCohere, line 368: Does it need to be tool specific? Couldn't we run 
`tool_auth_crud.delete_tool_auth(session, user_id, self.TOOL_ID)`
here?

src/backend/routers/auth.py
chantelle-cohere: gotcha, thank you! I will update to match your suggestion!

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+    session: DBSessionDep,
+    ctx: Context = Depends(get_context),
+):
+    """
+    Endpoint to delete Tool Authentication.
+
+    If completed, the corresponding ToolAuth for the requesting user is removed from the DB.
+
+    Args:
+        tool_id (str): ToolID to be deleted for the user. (eg. google_drive) Should be one of the values listed in the ToolName string enum class.
```
tianjing-li, line 335: Nit: `Tool ID` with whitespace

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+        request (Request): current Request object.
+        session (DBSessionDep): Database session.
+        ctx (Context): Context object.
+
+    Returns:
+        {"status": "success"} on successful deletion.
```
tianjing-li, line 341: To correspond to other DELETE endpoints, let's just return a `{}`. Usually we do this by adding a new pydantic schema with nothing in it. You can add a

```
class DeleteToolAuth(BaseModel):
    pass
```

pydantic schema and return that

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+
+    Raises:
+        HTTPException: If there was an error deleting the tool auth.
+    """
+
+    logger = ctx.get_logger()
+
+    def log_and_return_error(error_message: str):
```
tianjing-li, line 349: 2 things:

1. Let's rename this to `log_and_raise_exception` to better reflect what it performs
2. Let's move this out of the routers, and into the `services` folder, perhaps in the logger subfolder?

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+        log_and_return_error(f"Tool {tool_id} is not available in AVAILABLE_TOOLS.")
+
+    if tool.auth_implementation is None:
+        log_and_return_error(
+            f"Tool {tool.name} does not have an auth_implementation required for Tool Auth Deletion."
+        )
+
+    try:
+        tool_auth_service = tool.auth_implementation()
+        is_delete_tool_auth_successful = tool_auth_service.delete_tool_auth(
```
tianjing-li, line 380: Prefer: shorten the variable name to `deleted`, it should be clear enough

src/backend/routers/auth.py
```diff
@@ -317,3 +317,74 @@ def log_and_redirect_err(error_message: str):
...
+            session, user_id
+        )
+
+        if not is_delete_tool_auth_successful:
+            log_and_return_error("Error deleting Tool Auth.")
+
+    except Exception as e:
+        log_and_return_error(str(e))
+
+    return {"status": "success"}
```
tianjing-li, line 390: Here as mentioned above, let's return the pydantic schema `DeleteToolAuth`, you can check other delete endpoints for examples. This should return just a `{}`, the status code of 200 will be explicit enough that it was successful.

tianjing-li (changes requested):
Some small requests

src/backend/routers/auth.py
chantelle-cohere: Ah, I didn't know about this, thank you!

tianjing-li (approved):
lgtm

src/backend/routers/auth.py
chantelle-cohere: After the google drive tool/sync service session today, I thought it would make the most sense to put the tool auth deletion function in the `BaseToolAuthentication` class. Please let me know if there is a better place for it!

src/backend/services/logger/utils.py
```diff
@@ -20,3 +22,11 @@ def get_logger(self) -> BaseLogger:
         else:
             # Default to StructuredLogging
             return StructuredLogging(level, renderer)
+
+
+def log_and_raise_http_exception(logger: BaseLogger, error_message: str):
```
lusmoura, line 27: Nit: can we move this method to the logger class? This way we could call the function like `logger.log_and_raise(message)`

lusmoura (approved):
LGTM! Left one comment and would appreciate it if you could add a couple of unit tests

src/backend/tools/base.py
```diff
@@ -118,6 +118,14 @@ def retrieve_auth_token(
     def get_token(self, user_id: str, session: DBSessionDep) -> Optional[str]:
         return None
 
+    def delete_tool_auth(self, session: DBSessionDep, user_id: str) -> bool:
+        try:
+            tool_auth_crud.delete_tool_auth(session, user_id, self.TOOL_ID)
+            return True
+        except Exception as e:
+            logger.error(event=f"Error while deleting Tool Auth: {str(e)}")
```
BeatrixCohere, line 126: Should we cascade the error up?

src/backend/routers/auth.py
```diff
@@ -317,3 +319,68 @@ def log_and_redirect_err(error_message: str):
     response = RedirectResponse(redirect_uri)
 
     return response
+
+
+@router.delete("/tool/auth/{tool_id}")
+async def delete_tool_auth(
```
BeatrixCohere, line 325: Would be awesome to have some tests

BeatrixCohere (approved):

src/backend/services/logger/utils.py
chantelle-cohere: done!

src/backend/tools/base.py
chantelle-cohere: done!

src/backend/routers/auth.py
chantelle-cohere: Thank you for the reminder! I added unit tests for tool auth crud, and I will make a follow-up PR with integration tests, and tag you and @lusmoura!

PR #657, opened 2024-08-14
fix(toolkit): file upload for new conversations

knajjars:
Closes TLK-988

Couple of bug fixes/improvements:

- Fixes a bug when adding files to a new conversation
- Radio Component now accepts clicking on label
- Reset Files Params no longer reset tools


tomtobac (approved):
nice 🔥 

thanks for fixing the label on radio group, i was struggling the other day with that

PR #660, opened 2024-08-14
backend: prevent deleting Agent if does not belong to user-id

tianjing-li:

**AI Description**

This PR updates the agent deletion logic to allow only the owner to delete their agents. It also adds tests to validate the deletion behavior for both private and public agents.

## Summary
Previously, anyone could delete a public agent, but now only the owner can delete their agent, regardless of its privacy setting. This change enhances security and data privacy.

## Changes
- The `delete_agent` function in `src/backend/crud/agent.py` now only deletes an agent if it was created by the provided `user_id`.
- The `HTTPException` status code in `src/backend/routers/agent.py` has been updated from `500` to `401` to indicate unauthorized access when attempting to delete an agent not belonging to the user.
- New tests have been added in `src/backend/tests/unit/routers/test_agent.py` to verify the deletion of public and private agents and to ensure that agents not belonging to the user cannot be deleted.


EugeneLightsOn (approved):
Looks good to me

src/backend/crud/agent.py
```diff
@@ -277,23 +278,20 @@ def update_agent(
...
     Returns:
-      bool: True if the agent was deleted, False otherwise
+      bool: True if the Agent was deleted, False otherwise
     """
     agent_query = db.query(Agent).filter(Agent.id == agent_id)
     agent = agent_query.first()
 
-    if not agent:
-        return False
-
```
EugeneLightsOn, line 295: I think we need this check

EugeneLightsOn (commented):
Looks good to me. Just one thing—please restore the removed check for the agent's existence.

src/backend/crud/agent.py
```diff
@@ -277,23 +278,23 @@ def update_agent(
...
     """
     agent_query = db.query(Agent).filter(Agent.id == agent_id)
     agent = agent_query.first()
 
     if not agent:
         return False
 
-    if agent and agent.is_private and agent.user_id != user_id:
+    # Can only delete Agent created by user_id
+    if agent.user_id != user_id:
```
lusmoura, line 297: Instead of checking if the user_id is the same here, we could add the `user_id` to the query. 

`agent_query = db.query(Agent).filter(Agent.id == agent_id, Agent.user_id == user_id)`

tianjing-li: Good point, will do

lusmoura (approved):

src/backend/crud/agent.py
tianjing-li: done

PR #662, opened 2024-08-14
fix(toolkit): add loading screens and fix issues with code snippet

knajjars:
Add loading screens and fix issues with code snippet

**AI Description**

### DIFF
--- a/src/interfaces/assistants_web/package.json
+++ b/src/interfaces/assistants_web/package.json
@@ -9,7 +9,7 @@
     "yarn": "9999"
   },
   "scripts": {
-    "dev": "next dev --port 4444 --turbo",
+    "dev": "next dev --port 4444",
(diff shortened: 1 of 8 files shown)


src/interfaces/assistants_web/package.json
```diff
@@ -9,7 +9,7 @@
     "yarn": "9999"
   },
   "scripts": {
-    "dev": "next dev --port 4444 --turbo",
```
tomtobac, line 12: Which was the issue?

tomtobac (approved):

src/interfaces/assistants_web/package.json
knajjars: styles where not correct for syntax highlight =/ will resolve later when we get more time

PR #670, opened 2024-08-14
backend: compass file tool fix + generate FE client

scott-cohere:
Thank you for contributing to the Cohere Toolkit!

- [ ] **PR title**: "area: description"
  - Where "area" is whichever of interface, frontend, model, tools, backend, etc. is being modified. Use "docs: ..." for purely docs changes, "infra: ..." for CI changes.
  - Example: "deployment: add Azure model option"

- [ ] **PR message**: ***Delete this entire checklist*** and replace with
    - **Description:** a description of the change
    - **Issue:** the issue # it fixes, if applicable
    - **Dependencies:** any dependencies required for this change

- [ ] **Add tests and docs**: Please include testing and documentation for your changes
- [ ] **Lint and test**: Run `make lint` and `make run-tests`

**AI Description**

PR #681, opened 2024-08-16
feat: allow to delete an auth tool

tomtobac:
## Description

- Generate API client
- feat: allow to delete an auth tool.

Closes TLK-1001

### DIFF
--- a/src/interfaces/assistants_web/src/hooks/chat.ts
+++ b/src/interfaces/assistants_web/src/hooks/chat.ts
@@ -93,7 +93,7 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
   } = useFilesStore();
   const queryClient = useQueryClient();
 
-  const currentConversationId = id || composerFiles[0]?.conversation_id;
+  const currentConversationId = id || composerFiles[0]?.conversation_id || '';
 
   const [userMessage, setUserMessage] = useState('');
   const [isStreaming, setIsStreaming] = useState(false);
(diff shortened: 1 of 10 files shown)


src/interfaces/assistants_web/src/hooks/chat.ts
```diff
@@ -93,7 +93,7 @@ export const useChat = (config?: { onSend?: (msg: string) => void }) => {
   } = useFilesStore();
   const queryClient = useQueryClient();
 
-  const currentConversationId = id || composerFiles[0]?.conversation_id;
+  const currentConversationId = id || composerFiles[0]?.conversation_id || '';
```
knajjars, line 96: this will change soon https://github.com/cohere-ai/cohere-toolkit/pull/670

src/interfaces/assistants_web/src/hooks/files.ts
```diff
@@ -140,11 +142,14 @@ export const useFileActions = () => {
...
         newFileIds.push(uploadedFile.id);
         setParams({ fileIds: newFileIds });
-        addComposerFile(uploadedFile);
+        addComposerFile({ ...uploadedFile, user_id: userId });
       });
 
       if (!conversationId) {
-        setConversation({ id: uploadedFiles[0].conversation_id });
+        const newConversationId = uploadedFiles[0].conversation_id;
+        if (newConversationId) {
```
knajjars, line 150: probably good to wait for @scott-cohere changes to land before doing these changes

tomtobac: @knajjars This was due to the new API client generation; since it's already on main and this is just typescript, the actual code already looks like this.

PR #687, opened 2024-08-16
Add frontend redirect query param for tool auth

tianjing-li:
FE requirement that we need to be able to specify the redirect url back to the frontend, depending on where the user performs the initial tool auth call from. This allows the FE to generically send a state value to the OAuth call.


BeatrixCohere (approved):

EugeneLightsOn (approved):
Looks good to me

PR #697, opened 2024-08-19
fix(toolkit): integrate redirect URL for tool auth

knajjars:
Closes OS-2422

Integrates redirect URL for tool auth.

We currently have 3 different pages from where users can authenticate their GDrive. 
- `/settings`
- `/new`
- `/edit/:agent_id`

For the tool auth redirection we had to introduce deep linking state management, for which we use the `state` query param in the tool auth, where we pass a value of `frontend_redirect` which contains the URL the flow will redirect once it successfully gets the auth token. 

For the case when the user has state that we want to preserve, we can pass a hash of the JSON object for the state we want, in this case we use `atob` and `btoa` to encode and decode the state, which we will persist once the page loads with the value in the query params.


tomtobac (approved):
awesome 🔥

PR #721, opened 2024-08-21
fix: citations breaking leading markdown

tomtobac:
## Description

Some citations broke down the markdown elements due to the citation replacement. With the new `fixCitationsLeadingMarkdown`, we are checking if the citation could have the same opening tag after the citation; if so, we add it to the citation.

This fix only fixes leading markdown citations and only some elements, such: ` * or **

e.g.: 
```
Abra evolves into `Kadabra`
```

And the citation looks like
```
{
  text: '`Kadabra',
  start: 18,
  end: 26,
  document_ids: ['111111'],
},
```

We will update the citations to look like

```
{
  text: '`Kadabra`',
  start: 18,
  end: 27,
  document_ids: ['111111'],
},

``` 

Closes TLK-1033

### Screenshots

| Before | After |
|--------|--------|
| before | after |


src/interfaces/assistants_web/src/utils/citations.ts
```diff
@@ -11,6 +11,26 @@ export const fixMarkdownImagesInText = (text: string) => {
 
 const formatter = new Intl.ListFormat('en', { style: 'long', type: 'conjunction' });
 
+export const fixCitationsLeadingMarkdown = (citations: Citation[], originalText: string) => {
+  const citationsCopy = [...citations];
+  const markdownFixList = ['`', '*', '**'];
```
knajjars, line 16: we are not covering links, bullet points, etc. Right?

knajjars (approved):

src/interfaces/assistants_web/src/utils/citations.ts
tomtobac: Yes, not yet unless we see them

PR #725, opened 2024-08-21
feat: allow to download model generated files other than images

tomtobac:
## Description

It fixes citations, other than images, that are to be skipped for "encode". It checks if they have any output files and allows downloads.

Closes TLK-789

### Screenshots

image


src/interfaces/assistants_web/src/utils/file.ts
```diff
@@ -36,3 +36,30 @@ export const getFileUploadTimeEstimateInMs = (fileSizeInBytes: number) => {
...
   return ((3.51 * fileSizeInBytes) / 1000000 - 0.3) * 1000;
 };
+
+/**
+ * Encode text to base64.
+ * @param text
+ * @returns base64 encoded text
+ */
+export const decodeBase64 = (text: string) => {
+  return Buffer.from(text, 'base64').toString('utf-8');
```
abimacarmes, line 46: Should the docstring be decode rather than encode?

abimacarmes (approved):

PR #734, opened 2024-08-22
backend: Upgrade langchain and fix /langchain-chat endpoint

tianjing-li:

**AI Description**

This PR updates the `langchain dependency to version `0.2.14 and its related packages. It also updates the `ruff` package to version `0.6.0.

## Package Updates
- `langchain` is updated to version `0.2.14.
- `langchain-community` is updated to version `0.2.12.
- `langchain-core` is updated to version `0.2.34.
- `langchain-text-splitters` is updated to version `0.2.2.
- `ruff` is updated to version `0.6.0.

## Code Changes
- Added imports from `logger` module in `langchain.py`.
- Added error logging in `langchain.py` when a tool is not found or does not have the `to_langchain_tool` method implemented.
- Updated the error message in `chat.py` to fix a typo.
- Added headers, send timeout, and ping to the `langchain_chat_stream` function in `chat.py`.
- Updated the comment and logic for extracting content from `action.message_log` in `chat.py`.
- Updated the comment for the shape of the content variable in `chat.py`.
- Removed the `documents` field from the `StreamToolResult` class in `chat.py`.
- Updated the final output logic in `chat.py` to use the `output` field directly and removed the unused `citations` and `documents` fields.


src/backend/chat/custom/langchain.py
```diff
@@ -41,12 +44,16 @@ def chat(self, chat_request: LangchainChatRequest, **kwargs: Any) -> Any:
...
             tool = AVAILABLE_TOOLS.get(req_tool.name)
-            if tool:
+
+            if not tool:
+                logger.error(event=f"Failed to find tool {req_tool.name} during Langchain chat call.")
+
+            if not hasattr(tool.implementation, "to_langchain_tool"):
+                logger.error(event=f"Tool {req_tool.name} does not have to_langchain_tool method implemented.")
+
                 tools.append(tool.implementation().to_langchain_tool())
```
malexw, line 54: Is the logic here correct? As it is now, it looks like the tool will only be appended if it doesn't have `to_langchain_tool` implemented. Do you want `continue` statements after each of the `logger.error` lines?

This also seems like a good area for a unit test.

tianjing-li: You're right, fixed it by adding continues and moving the append code outside of the ifs (the latter error seems to be a copy paste issue)

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
tomtobac, line 1: mmm did we push it unintentionally?

tianjing-li:
No longer needed

PR #739, opened 2024-08-22
backend: Add Google Web Search tool

malexw:
Introduces a new tool for assistants to perform Google web searches, via the Custom Search Engine API. Requires the `google-api-python-client` package, which is included already.

**AI Description**

This pull request adds a new feature to perform Google web searches and retrieve relevant results. 

**Code Changes:**
- A new class `GoogleWebSearchSettings` is introduced in `settings.py`, inheriting from `BaseSettings` and `BaseModel`. It includes the `api_key` and `cse_id` fields, which are set from the configuration.
- The `ToolSettings` class in `settings.py` now includes a `google_web_search` field, with a default value of `GoogleWebSearchSettings()`.
- The `ToolName` enum in `tools.py` has been updated to include `Google_Web_Search`, with the appropriate display name, implementation, and parameter definitions.
- The `get_filtered_domains` method is added to the `GoogleWebSearch` class in `google_search.py` to retrieve filtered domains from agent tool metadata.
- The `call` method in `google_search.py` performs the actual Google web search using the provided query, filtered domains, and Google Custom Search Engine credentials. It returns a list of search results with titles, URLs, and snippets.
- Necessary imports and references have been added across the affected files to support the new Google web search functionality.

**Summary:**
The pull request enhances the system's capabilities by integrating Google web search functionality. This addition allows for performing web searches, handling search queries, and returning relevant search results. The changes include the necessary configuration, class definitions, and method implementations to facilitate this new feature.


src/backend/tools/google_search.py
```diff
@@ -0,0 +1,70 @@
...
+    NAME = "google_web_search"
+    API_KEY = Settings().tools.google_web_search.api_key
+    CSE_ID = Settings().tools.google_web_search.cse_id
+
+    def __init__(self):
+        self.client = build("customsearch", "v1", developerKey=self.API_KEY)
+
+    @classmethod
+    def is_available(cls) -> bool:
+        return cls.API_KEY is not None and cls.CSE_ID is not None
```
raphael-cohere, line 22: If you just do 
```
not cls.API_KEY  and not cls.CSE_ID
```
you will also guard against empty strings

src/backend/tools/google_search.py
```diff
@@ -0,0 +1,70 @@
...
+            if "domain" in artifact
+        ]
+
+    async def call(
+        self, parameters: dict, ctx: Any, session: DBSessionDep, **kwargs: Any
+    ) -> List[Dict[str, Any]]:
+        query = parameters.get("query", "")
+        cse = self.client.cse()
+        filtered_domains = self.get_filtered_domains(session, ctx)
+        if not filtered_domains:
```
raphael-cohere, line 53: You most likey don't need this check here:

```python
site_filters = [f"site:{domain}" for domain in filtered_domains]
response = cse.list(q=query, cx=self.CSE_ID, orTerms=site_filters).execute()
```
orTerms will be an empty list if there are no filtered_domains. If the library can't handle empty lists for or terms you can always shorten it to:

```python
response = cse.list(q=query, cx=self.CSE_ID, orTerms=site_filters or None).execute()
```

src/backend/tools/google_search.py
```diff
@@ -0,0 +1,70 @@
...
+    CSE_ID = Settings().tools.google_web_search.cse_id
+
+    def __init__(self):
+        self.client = build("customsearch", "v1", developerKey=self.API_KEY)
+
+    @classmethod
+    def is_available(cls) -> bool:
+        return cls.API_KEY is not None and cls.CSE_ID is not None
+
+    def get_filtered_domains(self, session: DBSessionDep, ctx: Context) -> list[str]:
```
raphael-cohere, line 24: this seems duplicated from tavily.py. you might want to extract this into a mixin:
```python
class DomainFilter:
  def get_filtered_domains(...)
  
class GoogleWebSearch(BaseTool, DomainFilter):
   pass

class TavilyInternetSearch(BaseTool, DomainFilter):
  pass
```

Or just extract it into a plain function somwhere and call it from both files.

raphael-cohere (approved):

src/backend/config/configuration.template.yaml
```diff
@@ -37,6 +37,9 @@ tools:
   google_drive:
     client_id:
     client_secret:
+  google_web_search:
```
tianjing-li, line 40: The api_key should be in the secrets template file.

For the `cse_id` I'm not too sure - if it's not sensitive info then you can keep it in the config file

src/backend/config/settings.py
```diff
@@ -177,6 +177,15 @@ class GDriveSettings(BaseSettings, BaseModel):
         ),
     )
 
+class GoogleWebSearchSettings(BaseSettings, BaseModel):
+    model_config = SETTINGS_CONFIG
+    api_key: Optional[str] = Field(
+        default=None, validation_alias=AliasChoices("API_KEY", "api_key")
```
tianjing-li, line 183: The alias determines the key for the `.env` (we keep this logic to make env users backwards compatible). I would change the name here to something more explicit, like `GOOGLE_SEARCH_API_KEY`

Same thing for the `CSE_ID`

src/backend/tools/google_search.py
tianjing-li: @raphael-cohere yup I'm planning to do this in a later PR, since Brave/Google will implement this. I can handle this

tianjing-li (changes requested):
Couple small asks

src/backend/tools/google_search.py
malexw: I need the opposite of that, so my options seemed to be:
```
return not(not cls.API_KEY and not cls.CSE_ID)
```
or
```
return bool(cls.API_KEY) and bool(cls.CSE_ID)
```
or
```
return cls.API_KEY is not None and len(cls.API_KEY) > 0 and cls.CSE_ID is not None and len(cls.CSE_ID) > 0
```

I also considered
```
return cls.API_KEY and cls.CSE_ID
```

but `type(cls.API_KEY and cls.CSE_ID) == "str"`

src/backend/tools/google_search.py
malexw: Good idea!

tianjing-li:
Closing this to move them all in one shot

PR #742, opened 2024-08-23
Brave web search tool

EugeneLightsOn:
- Brave web search tool
- AgentToolMetadata logic for the domain filtering using artifacts: 
```
"artifacts": [
            {
                "domain": "wikipedia.org"
            },
            {
                "domain": "test.org"
            }
        ]
```

**AI Description**

This PR introduces a new feature, the Brave Web Search tool, which allows users to perform web searches using the Brave Search API. 

The changes include:
- New `BraveWebSearch` class in `src/backend/tools/brave_web_search.py` that extends `BaseTool`.
- New `BraveClient` class in `src/backend/brave_search/brave_client.py` for interacting with the Brave Search API.
- New `SafeSearchTypes` enum for specifying the type of safe search: "off", "moderate", or "strict".
- New `WebSearchResponse` model for handling web search responses.
- New `BraveApiException` custom exception for handling Brave API errors.
- Updates to `settings.py`, `configuration.template.yaml`, `secrets.template.yaml`, and `tools.py` to include the new Brave Web Search tool.


src/backend/brave_search/brave_client.py
```diff
@@ -0,0 +1,302 @@
...
+class BraveClient:
+    BASE_URL = "https://api.search.brave.com/res/v1/"
+    WEB_SEARCH_API_ENDPOINT = "web/search"
+    MAX_RESULTS = 20
+    MAX_OFFSET = 9
+    WAIT_RETRY_SECONDS = 2
+    RETRY_ATTEMPTS = 2
+
+    def __init__(self, api_key: Optional[str] = None) -> None:
+        self.api_key = api_key if api_key else os.environ.get("BRAVE_API_KEY", None)
```
tianjing-li, line 51: For environment variables, we need to use the config `Settings()` class now, you can refer to the tavily client for an example.

The settings are class attributes instead of fetched during init, like so:

```
class TavilyInternetSearch(BaseTool):
    NAME = "web_search"
    TAVILY_API_KEY = Settings().tools.tavily.api_key
    POST_RERANK_MAX_RESULTS = 6

    def __init__(self):
        self.client = TavilyClient(api_key=self.TAVILY_API_KEY)
```

src/backend/brave_search/brave_client.py
```diff
@@ -0,0 +1,302 @@
...
+
+class BraveClient:
+    BASE_URL = "https://api.search.brave.com/res/v1/"
+    WEB_SEARCH_API_ENDPOINT = "web/search"
+    MAX_RESULTS = 20
+    MAX_OFFSET = 9
+    WAIT_RETRY_SECONDS = 2
+    RETRY_ATTEMPTS = 2
+
+    def __init__(self, api_key: Optional[str] = None) -> None:
```
tianjing-li, line 50: Rather than os.environ, can we use `Settings()` here to fetch the API key

src/backend/brave_search/brave_client.py
```diff
@@ -0,0 +1,302 @@
+import os
```
tianjing-li, line 1: Rather than create a folder in `backend`, I would suggest creating a `brave_search` folder in `backend/tools`, then move this to a `client.py` file, and your tool itself into a `tool.py` file

src/backend/config/configuration.template.yaml
```diff
@@ -37,6 +38,8 @@ tools:
   google_drive:
     client_id:
     client_secret:
+  brave_web_search:
```
tianjing-li, line 41: Let's remove this from config, and just keep it in the secrets file

EugeneLightsOn: Thank you

src/backend/config/tools.py
```diff
@@ -182,6 +184,23 @@ class ToolName(StrEnum):
...
+        display_name="Brave Web Search",
+        implementation=BraveWebSearch,
+        parameter_definitions={
+            "query": {
+                "description": "Query for retrieval.",
+                "type": "str",
+                "required": True,
+            }
+        },
+        is_visible=True,
```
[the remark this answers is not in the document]
EugeneLightsOn: Done

src/backend/brave_search/brave_client.py
EugeneLightsOn: Thanks

src/backend/brave_search/brave_client.py
EugeneLightsOn: I used os.environ in the Brave client as a fallback. The Settings are used in the tool to get the API key.

tianjing-li:
Closing this to move them all in one shot

PR #756, opened 2024-09-02
frontend: fix: remove `defaultProps` from `react-scroll-to-bottom`

lx-0:
This fixes the React warnings of package `react-scroll-to-bottom` regarding `defaultProps`. Package development seems to be stalled at the moment, therefore applying as patch.

Fix is automatically applied when running node package install.

Reference: [PR on `react-scroll-to-bottom`](https://github.com/compulim/react-scroll-to-bottom/pull/141)


lx-0:
See [PR 757](https://github.com/cohere-ai/cohere-toolkit/pull/757)

PR #784, opened 2024-09-20
backend: Remove metrics

tianjing-li:
- Per Cohere internal request, removing metrics code from the open source toolkit


pyproject.toml
```diff
@@ -104,10 +104,6 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.pyright]
-include = [
-  "src/backend/services/metrics.py", 
-  "src/backend/tools/google_drive/sync/actions/",
-  ]
 defineConstant = { DEBUG = true }
```
EugeneLightsOn, line 107: Do we need to check all the code? If so, I think we should set it to ./src/backend to check only the Python code.

EugeneLightsOn (commented):
Looks good to me, only one small question about pyright

pyproject.toml
tianjing-li: Let me try

tianjing-li: I had to revert this removal, even with the updated pyright checker it checks types across the entire backend - we have 900+ errors so there's no question that we don't have bandwidth to resolve them, I'll revert to this for now (which seems to resolve the error)

malexw (approved):
Nothing jumps out to me here as being a problem

PR #788, opened 2024-09-25
tools: Default auth to off and provide a better default DB path

malexw:
Two small changes based on issues that came up while I was testing `make first-run` recently:

- I don't think we actually want auth enabled by default for the demo, but maybe we should double-check. If we leave it enabled it adds an extra step to get things going, and it means we should be getting users to set up a JWT secret key during the setup process.

- Due to various dependencies and imports, the setup script actually requires a non-`None` value for the DB URL to be set. Since there's no config file by default on a fresh copy of the repo, we need to add a more realistic default value in `settings.py` 

**AI Description**

This pull request makes changes to the configuration and settings files in the backend of the application.

## Summary
The `src/backend/config/configuration.template.yaml` file has been updated to remove the `basic` authentication method from the `auth` section. This change suggests a shift away from basic authentication, potentially towards a more secure authentication mechanism.

The `src/backend/config/settings.py` file has been modified to set a default value for the `url` field in the `DatabaseSettings` class. The new default value is `postgresql+psycopg2://postgres:postgres@db:5432`, indicating a specific database connection string.

## Changes
- `src/backend/config/configuration.template.yaml`: Removed `basic` authentication method.
- `src/backend/config/settings.py`: Set a default value for the `url` field in the `DatabaseSettings` class.


tianjing-li (approved):
lgtm

PR #794, opened 2024-10-01
infra/docs: remove single Docker image setups

tianjing-li:
- Removes all single Docker image setups and deployments


malexw (approved):
🍻

thesethtruth:
What's the rationale for removing all single container builds? This also removed the only option for deploying to Azure. Thanks!

PR #802, opened 2024-10-06
frontend: Fall back to browser default clipboard behaviour if clipboard API is unavailable

### LINKED ISSUE #797: Unable to copy text from user/model in conversations
### What is the issue?

I have a local deployment of the cohere toolkit, and I am unable to copy text that I have previously sent in the chat, or text that the model has outputted. I do not get the text in my clipboard when I highlight it and copy it with control+c / right click + copy.  The same applies for the copy button next to the text, nothing happens.

### Additional information

Here is an image of the UI with the copy button I am referring to

image

malexw:
When the client is loaded over HTTP the clipboard is disabled. In that case, it's probably better to have the clipboard copy fix hook fall back to the default behaviour. In addition, let's have the copy button show a hint that the user should set up HTTPS to get the copy button working.

Fixes #797 

**AI Description**

This PR introduces a fix for a copy bug, ensuring that the Clipboard API is only used when available over HTTPS.

## Changes:
- Added a check to determine if the Clipboard API is available (`window.navigator.clipboard`) before attempting to use it.
- Updated the button label to display a message indicating that copying requires HTTPS if the Clipboard API is not available.
- Disabled the copy button when the Clipboard API is not available.
- Added a check to prevent the copy function from executing if the Clipboard API is not available.

### DIFF
--- a/src/interfaces/assistants_web/src/components/UI/CopyToClipboardButton.tsx
+++ b/src/interfaces/assistants_web/src/components/UI/CopyToClipboardButton.tsx
@@ -40,6 +40,9 @@ export const CopyToClipboardButton = forwardRef<
 ) {
   const [copied, setCopied] = useState(false);
 
+  const available = !!window?.navigator?.clipboard;
+  const buttonLabel = !available ? 'Copy (requires HTTPS)' : copied ? 'Copied!' : label;
+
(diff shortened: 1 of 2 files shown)


tianjing-li (approved):
lgtm

Issue #799, opened 2024-10-05
Switch to a better smooth scrolling to bottom package, use-stick-to-bottom


samdenty:
### What improvement(s) are you suggesting?

[npmjs.com/package/use-stick-to-bottom](https://npmjs.com/package/use-stick-to-bottom) is a new package designed for AI chatbot interfaces which is much better than the non-maintained react-scroll-to-bottom package.

It features a custom smooth scrolling algorithm with a velocity-based spring animation so when new content jumps into view, it adjusts the velocity dynamically based on how much more new content was added

PR #803, opened 2024-10-06
feat: switch to use-stick-to-bottom

samdenty:
This closes https://github.com/cohere-ai/cohere-toolkit/issues/799.

At stackblitz.com with the recent release of [bolt.new](https://bolt.new), we created a new package [npmjs.com/package/use-stick-to-bottom](https://npmjs.com/package/use-stick-to-bottom) designed for AI chatbot interfaces which is much better than the non-maintained react-scroll-to-bottom package.

It features a custom smooth scrolling algorithm with a velocity-based spring animation so when new content jumps into view, it adjusts the velocity dynamically based on how much more new content was added resulting in a much less jumpy animation than the package you were using before

I also see you had to do some patch-package hacks - this new package is designed to work with the latest react without any errors


tianjing-li (approved):
This is great, did you have a chance to test the functionality?

samdenty:
> This is great, did you have a chance to test the functionality?

I haven't tested no because I don't know how to, but maybe someone on your team wants to check it works. The hook itself is solid we've done extensive back and forth

tianjing-li:
@samdenty I'll try to test this out when time permits

PR #814, opened 2024-10-16
Improve agent creation flow

EugeneLightsOn:
Create deployments and models during the agent creation flow if they are not present in the database but have configurations in the codebase.

**AI Description**

This PR introduces new functions and updates existing ones to enhance the deployment and model management system.

- **New Functions:**
  - `find_config_by_deployment_id` and `find_config_by_deployment_name` in `src/backend/config/deployments.py` enable configuration retrieval by deployment ID or name.
  - `create_deployment_by_config` in `src/backend/crud/deployment.py` facilitates deployment creation from a configuration.
  - `get_model_by_name` in `src/backend/crud/model.py` retrieves a model by its name.
  - `create_model_by_config` in `src/backend/crud/model.py` creates a new model based on a configuration.

- **Updated Functions:**
  - `validate_deployment_model` in `src/backend/services/request_validators.py` now checks deployment config settings and creates deployments and models if they are not found.
  - `test_create_agent_deployment_not_in_db` in `src/backend/tests/unit/routers/test_agent.py` tests agent creation with a deployment not in the database.
  - `test_streaming_chat_with_tools_not_in_agent_tools` and `test_streaming_chat_with_agent_tools_and_empty_request_tools` in `src/backend/tests/unit/routers/test_chat.py` have been updated to use the "tavily_web_search" tool.


src/backend/crud/deployment.py
```diff
@@ -184,3 +190,33 @@ def delete_deployment(db: Session, deployment_id: str) -> None:
...
+        deployment_config (DeploymentSchema): Deployment config.
+
+    Returns:
+        Deployment: Created deployment.
+    """
+    deployment = Deployment(
+        name=deployment_config.name,
+        description="",
+        default_deployment_config= {
+                env_var: os.environ.get(env_var, "")
```
tianjing-li, line 212: This would create a Deployment when an env variable could be unset with the `""` default, open to how to handle that scenario. Maybe just iterate through the variables before creating the deployment and check they're not None

tianjing-li (approved):
Small change requested

src/backend/crud/deployment.py
EugeneLightsOn: I don't think we need to do this in the agent creation flow. I believe this check should be part of the Chat flow, and it’s already covered by the models. What do you think?

PR #815, opened 2024-10-16
backend: Add death loop detection

tianjing-li:
- Imports Death loop from eno


EugeneLightsOn (approved):
Looks good to me. We need to re-sync it again after the Eno experiments are done.

PR #816, opened 2024-10-16
Revert "feat: switch to use-stick-to-bottom"

tianjing-li:
Reverts cohere-ai/cohere-toolkit#803


tianjing-li:
Hi @samdenty we were running into an issue where the text box was levitating image 

I didn't have any bandwidth to look into resolving this, so I reverted it for now. No problems with adding the new lib but would need to fix that issue

malexw (approved):

PR #818, opened 2024-10-17
Add Lightson Codeowners

tianjing-li:
Adds more Lightson codeowners for the frontend portions to reduce friction


EugeneLightsOn (approved):
Looks good to me

PR #819, opened 2024-10-17
frontend: Fix Agent model being set

tianjing-li:
**AI Description**

This pull request updates the `useState` hook in the `ConfigStep` component to use the `fields.model` value instead of the `DEFAULT_AGENT_MODEL` constant.

- The `useState` hook now initializes the `selectedValue` state with the value of `fields.model`.
- This change allows the component to dynamically set the initial value of the `selectedValue` state based on the `fields` prop, providing more flexibility in setting the default agent model.


malexw (approved):
Good fix!

PR #833, opened 2024-11-07
infra: Enable integration tests for all PRs and fix broken tests

malexw:
Resolves an issue that was causing two broken integration tests. (And in fact, the feature was broken but I don't think anyone noticed).

We also now can run integration tests for all PRs. 

**AI Description**

This PR introduces changes to the backend integration tests and the conversation service.

## Changes
- The `backend_integration_tests.yml` file now includes a new trigger for pull requests.
- The `conversation.py` file has been updated to import the `uuid` module.
- A new function, `create_message`, has been added to the `conversation.py` file. This function is responsible for creating a message with specified parameters, including session, chat request, conversation ID, user ID, and message agent.
- The `generate_conversation_title` function in the `conversation.py` file now calls the `create_message` function to generate a response message.


EugeneLightsOn (approved):
Nice PR

EugeneLightsOn (commented):
Alex but as I see there is another error which is not related to the PR.

.github/workflows/backend_integration_tests.yml
```diff
@@ -3,6 +3,7 @@ name: Backend - Integration tests
 on:
   push:
     branches: [main]
+  pull_request: {}
   merge_group: {}
```
tianjing-li, line 7: Let's remove merge_group now, since this runs on each commit

tianjing-li:
Could you follow up with Luisa if any keys need to be added for huggingface, otherwise we can disable the failing tests

malexw:
> Could you follow up with Luisa if any keys need to be added for huggingface, otherwise we can disable the failing tests

Will do

tianjing-li (approved):
Good to go after initial comments

PR #838, opened 2024-11-11
***Tools***: Add Gmail support

scottmx81:
**Description:** This PR adds a Gmail tool that allows users to access emails from their Gmail account.

**Dependencies:** This features requires the user to have a Gmail account, and for the Google Cloud Console to have Gmail API access enabled. This PR does not add any new Python dependencies to `pyproject.toml` because it uses the same Python SDK that is already in the Toolkit for accessing Google Drive.

**AI Description**

### DIFF
--- a/docs/custom_tool_guides/gmail.md
+++ b/docs/custom_tool_guides/gmail.md
@@ -1,0 +1,82 @@
+# Gmail Tool Setup
+
+To set up the Gmail tool you will need to configure API access in Google Cloud Console.
+
+Follow the steps below to set it up:
+
+## 1. Create Project in Google Cloud Console 
+
+Head to the [Google Cloud Console](https://console.cloud.google.com/) and create a new project.
+After creating the app, you will see the `APIs & Services` section. Under `Enabled APIs & services`, enable the
+Gmail API.
+
+## 2. Configure OAuth Consent Screen
+
+Before you can generate the API credentials, you must first configure the OAuth consent screen.
+
+You will need to configure the `Application home page` and `Authorized domain 1` fields, with a URL and domain that
+point to where you are running the Toolkit. If you are running it in a local environment, Ngrok can be used as a proxy
+to access the to the Toolkit in local. Using `localhost` is not accepted value for these fields.
+
+If you choose to use Ngrok, you can start it with:
+
+`ngrok http -domain <your_custom_domain>.ngrok.dev 8000`
+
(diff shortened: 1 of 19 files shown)


docs/custom_tool_guides/gmail.md
```diff
@@ -0,0 +1,82 @@
...
+`BASE_AGENT_EXCLUDED_TOOLS` list. By default, the Gmail Tool is disabled for the Base Agent.
+
+```typescript
+export const BASE_AGENT_EXCLUDED_TOOLS = [];
+```
+
+## 5. Run the Backend and Frontend
+
+run next command to start the backend and frontend:
+
```
tianjing-li, line 79: Great docs

src/backend/tools/gmail/auth.py
```diff
@@ -0,0 +1,138 @@
...
+class GmailAuth(BaseToolAuthentication, ToolAuthenticationCacheMixin):
+    TOOL_ID = GMAIL_TOOL_ID
+    AUTH_ENDPOINT = "https://accounts.google.com/o/oauth2/v2/auth"
+    TOKEN_ENDPOINT = "https://oauth2.googleapis.com/token"
+    DEFAULT_USER_SCOPES = ['https://www.googleapis.com/auth/gmail.readonly']
+
+    def __init__(self):
+        super().__init__()
+
+        settings = Settings()
```
tianjing-li, line 31: The Settings() usage can be updated now to use `.get()`

src/backend/tools/gmail/tool.py
```diff
@@ -0,0 +1,78 @@
...
+    Tool that searches Gmail for messages and files based on a query.
+    """
+
+    ID = GMAIL_TOOL_ID
+    CLIENT_ID = ""
+    CLIENT_SECRET = ""
+
+    @classmethod
+    def is_available(cls) -> bool:
+        settings = Settings()
```
tianjing-li, line 26: Same thing here, let's update Settings() to use `.get()`

tianjing-li: Please also move the attribute setting out of `is_available` and directly onto the class attribute, then only check that they're not None in `is_available`

tianjing-li (changes requested):
Some v minor changes then good to go!

tianjing-li (approved):

PR #839, opened 2024-11-12
frontend: Allow toggling tools for non-default agents

tianjing-li:
Allow toggling tools for custom Assistants too

**AI Description**

This pull request makes changes to the `DataSourceMenu.tsx`, `AssistantTools.tsx`, and `Welcome.tsx` components.

## Summary
The changes remove the `checkIsBaseAgent` function from the imports and usage in the `DataSourceMenu.tsx` and `AssistantTools.tsx` components. Instead, the `isBaseAgent` variable is replaced with a `paramsTools` variable, which is used to determine the number of tools and the state of the switch component.

## Changes
- The `checkIsBaseAgent` function is removed from the imports in `DataSourceMenu.tsx` and `AssistantTools.tsx`.
- In `DataSourceMenu.tsx`, the `isBaseAgent` variable is replaced with a `paramsTools` variable, which is used to determine the number of tools and the state of the switch component.
- In `AssistantTools.tsx`, the `disabled` prop of the `ToggleCard` component is removed, and the `onToggle` prop is updated to use the `name` property of the tool.
- In `Welcome.tsx`, the `isBaseAgent` variable is removed, and the `Toggle Tools On/Off` text is always displayed.

### DIFF
--- a/src/interfaces/assistants_web/src/components/MessagingContainer/AssistantTools.tsx
+++ b/src/interfaces/assistants_web/src/components/MessagingContainer/AssistantTools.tsx
@@ -7,7 +7,7 @@ import { WelcomeGuideTooltip } from '@/components/MessagingContainer';
 import { Button, Icon, Text, ToggleCard } from '@/components/UI';
 import { useAvailableTools } from '@/hooks';
 import { useParamsStore } from '@/stores';
-import { checkIsBaseAgent, cn, getToolIcon } from '@/utils';
+import { cn, getToolIcon } from '@/utils';
 
 /**
  * @description Tools for the assistant to use in the conversation.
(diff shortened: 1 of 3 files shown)


src/interfaces/assistants_web/src/components/MessagingContainer/AssistantTools.tsx
```diff
@@ -41,18 +41,16 @@ export const AssistantTools: React.FC<{
...
                 <ToggleCard
                   key={name}
-                  disabled={disabled}
                   errorMessage={error_message}
                   checked={checked}
                   label={display_name ?? name ?? ''}
                   icon={getToolIcon(name)}
                   description={description ?? ''}
-                  onToggle={(checked) => handleToggle(name ?? '', checked)}
+                  onToggle={(checked) => handleToggle(name!, checked)}
```
malexw, line 53: Obviously you don't have to be the one to fix this, but it's a bit surprising and challenging that `name` can be null in `ToolDefinition`

malexw (approved):

knajjars (approved):

src/interfaces/assistants_web/src/components/MessagingContainer/AssistantTools.tsx
tianjing-li: Yeah there are a lot of fallbacks for various attributes in the client-side. Something we could refine by adding better type checking

PR #847, opened 2024-11-18
chore: Fix Groups/UserGroupsAssociation table overlap issue

tianjing-li:
These relationship overlap but in a harmless way, we explicitely add the overlap field here in the sqlalchemy models to silence the warning. 
**AI Description**

This PR updates the `user_associations` relationship in the `Group` and `UserGroupAssociation` classes to include the `overlaps` parameter.

- In `Group`, the `user_associations` relationship is updated to include the `overlaps` parameter with the value `"groups,users"`.
- In `UserGroupAssociation`, the `group` relationship is updated to include the `overlaps` parameter with the value `"groups,user_associations,users"`.

### DIFF
--- a/src/backend/database_models/group.py
+++ b/src/backend/database_models/group.py
@@ -10,6 +10,6 @@ class Group(Base):
     display_name: Mapped[str] = mapped_column(String)
 
     users = relationship("User", secondary="user_group", backref="groups")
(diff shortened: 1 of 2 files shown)


EugeneLightsOn (approved):
Looks good to me

PR #849, opened 2024-11-19
Backend: Check tool parameters generated by model

EugeneLightsOn:
The metaclass of the tool's base class decorates the call method of the tools to validate the passed parameters.
**AI Description**

This pull request introduces a new `ParametersCheckingMeta` metaclass to the `BaseTool` class, which automatically decorates the `call` method with a parameter checker. This checker validates the parameters passed to the `call` method against the tool's parameter definitions, ensuring that required parameters are provided and that the types of the parameters match the expected types.

The parameter checker is implemented using two new functions: `check_type` and `check_tool_parameters`. The `check_type` function recursively checks the type of a value against a given type description, while the `check_tool_parameters` function decorates the `call` method and performs the actual parameter validation.

The `ParametersCheckingMeta` metaclass is used as the metaclass for the `BaseTool` class, ensuring that all subclasses of `BaseTool` have their `call` method decorated with the parameter checker.

- The `check_type` function is added to check the type of a value against a given type description.
- The `check_tool_parameters` function is added to decorate the `call` method and perform parameter validation.
- The `ParametersCheckingMeta` metaclass is introduced to automatically decorate the `call` method with the parameter checker.
- The `BaseTool` class is modified to use the `ParametersCheckingMeta` metaclass.
- The `ToolDefinition` class is updated to include a `get_tool_definition` method, which returns a `ToolDefinition` object.
- The `mock_get_available_tools` function is modified to use the `get_tool_definition` method to get the tool definition for a given tool.
- The `test_async_call_tools_success`, `test_async_call_tools_failure`, `test_async_call_tools_timeout`, and `test_async_call_tools_failure_and_success` functions are updated to use the `get_tool_definition` method to get the tool definition for the `MockCalculator` and `MockWebScrape` tools.
- Three new test functions are added: `test_tools_params_checker_invalid_param`, `test_tools_params_checker_invalid_param_type`, and `test_tools_params_checker_required_param_empty`. These functions test the parameter checker by passing invalid parameters to the `call` method and asserting that the expected errors are raised.


src/backend/tools/base.py
```diff
@@ -31,7 +31,88 @@ class ToolError(BaseModel, extra="allow"):
     text: str
     details: str = ""
 
-class BaseTool():
+def check_type(param_value, type_description: str) -> bool:
+    try:
+        # Convert the type string into a type object
+        expected_type = eval(type_description)
```
tianjing-li, line 37: Can eval() be exploited somehow?

src/backend/tools/base.py
```diff
@@ -31,7 +31,88 @@ class ToolError(BaseModel, extra="allow"):
     text: str
     details: str = ""
 
-class BaseTool():
+def check_type(param_value, type_description: str) -> bool:
```
tianjing-li, line 34: Nit: can we move check_type, check_type_recursive and check_tool_parameters into the tool.utils folder, then keep the metaclass here

src/backend/tools/base.py
```diff
@@ -31,7 +31,88 @@ class ToolError(BaseModel, extra="allow"):
...
+                        )
+
+            return func(self, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+class ParametersCheckingMeta(type):
```
tianjing-li, line 104: nit: perhaps `ParametersValidationMeta`

tianjing-li (approved):

src/backend/tools/base.py
EugeneLightsOn: We use only developers settings here

PR #859, opened 2024-11-25
backend: Filter Tool Auth client access token

tianjing-li:
Only returns access token for Google Drive tool auth now, by default.

Adds some new unit tests

**AI Description**

This PR introduces a new package, `fakeredis`, and its associated dependencies, including `sortedcontainers`. The package is used for testing purposes and provides a Python implementation of the Redis API.

The PR also includes changes to the `tool.py` and `tool.test.py` files, which add a new field, `should_return_token`, to the `ToolDefinition` class. This field is used to determine whether an access token should be returned to the client when required by the frontend.

The `tool.test.py` file has been updated to include new tests for the `list_tools` function, which now checks for the presence of the `should_return_token` field in the tool definitions.


EugeneLightsOn (approved):
Looks good to me.

PR #862, opened 2024-11-29
backend: Filter out invalid message text when generating chat history

tianjing-li:
Thanks @EugeneLightsOn  for pointing out fix area

Right now if a chat stream fails after creating the initial Chatbot response message, it creates an empty chat in chat_history, which the chat model does not accept. This filters them out.

Ideally we go back and delete those previous messages, but I didn't want to introduce potentially janky logic.

**AI Description**

This PR introduces several changes to the chat functionality, enhancing the user experience and improving the backend logic.

## Summary
The `create_chat_history` function in `src/backend/services/chat.py` now filters out the user's most recent message and any empty messages when generating chat history. This ensures that only relevant messages are included in the history, providing a more accurate context for the chat SDK call.

Additionally, a new function `is_final_event` is added to `src/backend/chat/custom/custom.py`, which appears to be related to event handling. However, the purpose of this function is not entirely clear from the provided code snippet.

## Changes
  - The `create_chat_history` function now filters out the user's most recent message and any empty messages.
  - A new function `is_final_event` is introduced, which seems to be related to event handling.
- **src/backend/chat/custom/tool_calls.py**:
  - A new function `_call_tool_async` is added, but its purpose is not evident from the provided code.


EugeneLightsOn (approved):
looks good to me

PR #866, opened 2024-12-03
backend: Update make first-run to read existing configuration variables

scottmx81:
This PR modifies the behaviour of running the `make first-run` command so that it reads the variables from any existing `.env`, `configuration.yaml` or `secrets.yaml` files.

To do this, I have made use of the existing `ENV_YAML_CONFIG_MAPPING` variable, but I have added a `convert_yaml_to_secrets` function, which does the opposite of the existing `convert_secrets_to_yaml` function.

In the `backend/cli/main.py` script, it collects the answers from the users, and stores everything in a dict, using keys that correspond to the env var names in the `.env` file. In order for there to be minimal changes to the CLI script and how it writes out the variables, I've kept that existing way of working with the `secrets` dict in place, and use the new `convert_yaml_to_secrets` function to get the nested config structure from the yaml file into the "flat" structure in the `secrets` dict in `main.py`.

If any variables have been set that are recognized (i.e. they exist as keys in `ENV_YAML_CONFIG_MAPPING`), the user will not be prompted to enter the values again when running `make first-run`. This only works with the keys defined in the mapping, and there is one that is currently commented out in the code in the `main` branch: `NEXT_PUBLIC_API_HOSTNAME`. At the time of opening this PR, that variable does not get preserved if it has already been set in the `.env` file when running `make first-run`. I don't fully understand why that is commented out and does not have a YAML equivalent. If we could go ahead and make a YAML equivalent, it would fix it here.

I have also modified the behaviour of `make first-run` so that it prompts the user to overwrite the existing `configuration.yaml` and / or `secrets.yaml` files if they can not be read by `read_yaml` function. If it raises a `yaml.scanner.ScannerError` exception, the user will see one of the following prompts:

`Your existing configuration file is invalid. Overwrite? (y/N):`

or

`Your existing secrets file is invalid. Overwrite? (y/N):`

Pressing `n` will immediately exit the script. If the user types `y`, it will delete the existing `configuration.yaml` or `secrets.yaml` file that raised the exception.

I have tested the script when `configuration.yaml` and `secrets.yaml` are empty directories instead of files, and it replaces them with correct files. There was a fix for handling that edge case in another PR , but I have tested that the changes here still work as I would expect when those are directories instead of files due to a bug.

I had to make the `AVAILABLE_MODEL_DEPLOYMENTS` imports local because the definition of those variables indirectly uses settings, and would throw an exception if the `configuration.yaml` or `secrets.yaml` files were invalid. In other words, even without the new call to `read_yaml` that I have added in `cli/main.py`, running `make first-run` would fail with a stack trace if the files were invalid. The call to read the yaml file is buried deep, and the import of `AVAILABLE_MODEL_DEPLOYMENTS` itself would raise the exceptions, not when it was used, so the import itself would have to be wrapped in the try/except block. Changing it to local import in `cli/main.py` file allows the `read_yaml` function call that I've added to be wrapped in the try/except block, and to show the user the more friendly invalid config prompts, rather than seeing the stack trace before the `start()` function is ever called.

**AI Description**

This PR introduces a new approach to handling configuration and secrets files, aiming to streamline the setup process and enhance user experience. It adds functionality to validate and process existing YAML configuration files, ensuring compatibility and avoiding potential issues.

## Summary of Changes
- **Configuration and Secrets Handling:** The PR introduces a new `dotenv_values` function to retrieve secrets, replacing the previous approach of using an empty dictionary.
- **YAML Configuration Processing:** A new function, `process_existing_yaml_config`, is added to validate and process YAML configuration files. It checks for file existence, handles potential errors, and updates secrets accordingly.
- **Secret Retrieval and Prompting:** The `core_env_var_prompt` function now retrieves secrets from the `secrets` dictionary, prompting the user for input only if a secret is missing.
- **Community Feature Setup:** The `community_tools_prompt` function now checks the `USE_COMMUNITY_FEATURES` flag in the `secrets` dictionary before prompting the user to set up community features.
- **Build Target Selection:** The `build_target_prompt` function now retrieves the build target from the `secrets` dictionary, prompting for input only if it is not already set.

## Detailed Changes
- **src/backend/cli/main.py:**
  - Imports `dotenv_values` from `dotenv` and additional constants.
  - Adds `overwrite_config_prompt` and `overwrite_secrets_prompt` to the list of prompts.
  - Initializes `secrets` using `dotenv_values()` and processes YAML configuration files using `process_existing_yaml_config`.
  - Imports `AVAILABLE_MODEL_DEPLOYMENTS` and `COMMUNITY_MODEL_DEPLOYMENTS` after setting up the environment.
- **src/backend/cli/prompts.py:**
  - Modifies `core_env_var_prompt` to retrieve secrets from the `secrets` dictionary and prompt for input only if a secret is missing.
  - Renames `core_env_var_prompt` to `overwrite_config_prompt` and `overwrite_secrets_prompt`, prompting the user to overwrite invalid configuration and secrets files.
  - Updates `community_tools_prompt` to check the `USE_COMMUNITY_FEATURES` flag in the `secrets` dictionary before prompting for community feature setup.
  - Enhances `deployment_prompt` and `tool_prompt` to retrieve values from the `secrets` dictionary, prompting for input only if a value is missing.
  - Modifies `build_target_prompt` to retrieve the build target from the `secrets` dictionary, prompting for input only if it is not set.
- **src/backend/cli/utils.py:**
  - Adds imports for `sys`, `Path`, and `yaml`.
  - Introduces `process_existing_yaml_config` to validate and process YAML configuration files, handling file existence, errors, and secret updates.
  - Implements `convert_yaml_to_secrets` to convert YAML dictionary values to secrets, supporting nested values.


src/backend/cli/utils.py
```diff
@@ -12,6 +23,43 @@ def show_welcome_message():
...
+        try:
+            yaml_config = read_yaml(path)
+        except yaml.scanner.ScannerError:
+            if prompt():
+                yaml_config = {}
+                _path.unlink()
+            else:
+                sys.exit(1)
+
+        secrets.update(convert_yaml_to_secrets(yaml_config))
```
tianjing-li, line 39: Just want to make sure I understand the logic:

1. Earlier we load the `.env` values with `dotenv_values()`
2. Then we call this process function, which loads the YAML file (for either config/secrets)
3. YAML file is read and loaded
4. Iterate through each key of the .env <-> yaml mapping and if the value exists in the YAML file, we overwrite it for the secrets dictionary

tianjing-li (approved):
Looks good!

src/backend/cli/utils.py
scottmx81: @tianjing-li yes this is right. The YAML values will overwrite any values read from the `.env` file first. Is this OK?

tianjing-li: Yep sounds great

PR #873, opened 2024-12-06
frontend: Add special handling for HTTP errors when trying to begin a streaming chat

malexw:
While there was some handling for HTTP errors already, the default `onOpen` handler we were using with the `fetch-event-source` ended up swallowing HTTP errors and throwing its own that hid the underlying problem. In addition, a big try/catch wrapper around the `streamConverse` call was further masking error messages and potentially replacing some detailed error messages with more generic messages.

**AI Description**

This pull request introduces changes to the `CohereClient` class and the `useChat` function, along with a minor adjustment to the `MessageContent` component.

## **CohereClient Class:**
- The `onopen` property now features an asynchronous function that handles the response. If the response status is not 200 and the content type includes 'application/json', the function awaits the JSON response, catches any errors, and throws a `CohereNetworkError` with the error message and status.

## **useChat Function:**
- The `try-catch` block surrounding the `streamConverse` call has been removed, simplifying error handling.
- The `streamConverse` call now includes an `onError` callback, which handles errors and updates the conversation state accordingly.
- The `onError` callback distinguishes between `CohereNetworkError` and `StreamError`, setting error messages and updating the conversation accordingly.
- The `setConversation` function is called with the final message when the stream ends, and the `setStreamingMessage` function is set to null.

## **MessageContent Component:**
- The `button` element's class has been updated from `underline` to `ml-2 underline`.


tianjing-li (approved):
This is amazing. Would it be complex to add this behaviour to coral_web as well?

malexw:
> This is amazing. Would it be complex to add this behaviour to coral_web as well?

Yup - added now

PR #888, opened 2024-12-30
chore: Check config when running `make dev`

tianjing-li:
**AI Description**

This PR introduces changes to the `Makefile`, primarily focusing on the `check-config` target and related tasks.

## Summary
The `check-config` target has been split into two separate targets: `check-config-install` and `check-config`. The former now handles the installation process, while the latter is responsible for running the configuration check. This separation enhances clarity and modularity in the build process.

## Changes
- Added a new target `check-config-install` to handle the installation process.
- Modified the `check-config` target to only run the configuration check.
- Added a new line `make check-config` to the `dev` target, ensuring the configuration check is executed during development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,9 @@
 # Build & Run
 .PHONY: dev
 dev:
+	make check-config
 	make -j 2 watch up
-
+	
 .PHONY: watch
 watch:
 	@docker compose watch --no-up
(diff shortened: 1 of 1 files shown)


EugeneLightsOn (approved):
Looks good to me

Makefile
```diff
@@ -2,8 +2,9 @@
 # Build & Run
 .PHONY: dev
 dev:
+	make check-config
 	make -j 2 watch up
-
+
```
ezawadski, line 7: Personally I would clean up this white space.

ezawadski (approved):

Makefile
tianjing-li: Done

tianjing-li: I just took a look, it might be GH highlighting but it's not an extra newline, we still retain one newline between commands

PR #890, opened 2025-01-03
chore: Add coral_web client generation CI step

tianjing-li:
Blocked: waiting on Coral web fixes

**AI Description**

This pull request introduces several new steps to the workflow, enhancing the build process and ensuring the client is up-to-date.

- A new step, `Generate client`, is added to the workflow, which runs the command `npm run generate:client && npm run format:write`. This step generates the client code, ensuring it is up-to-date.
- Another step, `Check generated client`, is introduced to verify the generated client. It uses `git diff --exit-code ./src/cohere-client/generated` to compare the generated client with the expected version. If they differ, an error message is displayed, prompting regeneration and commitment.


EugeneLightsOn (approved):

EugeneLightsOn (commented):
It's a good idea, Thank you TJ

PR #892, opened 2025-01-07
[WIP] chore: Remove yarn 9999 in package.json 

tianjing-li:
This prevented dependabot runs, and was not necessary. Removing yarn and pnpm versions from the package.json engines
**AI Description**

This pull request updates the `package.json` files for the `src/interfaces/assistants_web` and `src/interfaces/coral_web` directories.

## Changes
- The `package.json` files for both directories have been updated to use a specific version of `pnpm` (`^9.10.0`) instead of the previous generic version (`9999`).
- The `yarn` dependency has been removed from both `package.json` files.

### DIFF
--- a/src/interfaces/assistants_web/package-lock.json
+++ b/src/interfaces/assistants_web/package-lock.json
@@ -109,24 +109,13 @@
       },
       "engines": {
         "node": ">=20.12.2",
-        "npm": ">=9",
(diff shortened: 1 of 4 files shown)


EugeneLightsOn (approved):
Thank you TJ

tianjing-li:
Resolved itself now

PR #901, opened 2025-01-09
frontend: added temperature gauge to assistant form

ezawadski:
Added the Slider Component from the Coral Web interface to the Assistants Web interface.

Added Temperature Setting to Assistant Create and Update forms. This setting is then used when chatting with the assistant.

**AI Description**

This PR introduces a new `temperature` parameter to the `AgentSettingsForm` component and its associated types, allowing users to set the temperature for agents. The `temperature` parameter is now included in the `CreateAgentSettingsFields` and `UpdateAgentSettingsFields` types, as well as in the `ConfigStep` component.

- The `temperature` parameter is added to the `AgentSettingsForm` component's `ConfigStep` component, allowing users to set the temperature for agents.
- The `temperature` parameter is included in the `CreateAgentSettingsFields` and `UpdateAgentSettingsFields` types, ensuring that the temperature is considered when creating or updating agents.
- The `temperature` parameter is set to a default value of `0.3` in the `DEFAULT_AGENT_TEMPERATURE` constant.
- The `temperature` parameter is added to the `Chat.tsx` component, enabling the setting of the temperature for the agent.
- The `temperature` parameter is included in the `UpdateAgent.tsx` component, allowing the temperature to be updated for the agent.
- The `temperature` parameter is added to the `CreateAgent.tsx` component, enabling the setting of the temperature for the agent during creation.
- The `temperature` parameter is included in the `paramsSlice.ts` file, ensuring that the temperature is considered when creating or updating agents.

### DIFF
--- a/src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
+++ b/src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
@@ -14,7 +14,7 @@ type Props = {
 
 export const ConfigStep: React.FC<Props> = ({ fields, setFields }) => {
   const [selectedModelValue, setSelectedModelValue] = useState<string | undefined>(fields.model);
-  const [selectedTempValue, setSelectedTempValue] = useState<number | undefined>(
+  const [selectedTemperatureValue, setSelectedTemperatureValue] = useState<number | undefined>(
     fields.temperature
   );
 
@@ -41,10 +41,10 @@ export const ConfigStep: React.FC<Props> = ({ fields, setFields }) => {
         min={0}
         max={1.0}
         step={0.1}
-        value={selectedTempValue || 0}
+        value={selectedTemperatureValue || 0}
         onChange={(temperature) => {
           setFields({ ...fields, temperature: temperature });
-          setSelectedTempValue(temperature);
+          setSelectedTemperatureValue(temperature);
         }}
       ></Slider>
     </div>
--- a/src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
+++ b/src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
@@ -3,7 +3,7 @@
 import { useState } from 'react';
 
 import { AgentSettingsFields } from '@/components/AgentSettingsForm';
-import { Dropdown } from '@/components/UI';
+import { Dropdown, Slider } from '@/components/UI';
 import { useListAllDeployments } from '@/hooks';
 
 type Props = {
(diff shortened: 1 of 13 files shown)


src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
```diff
@@ -13,7 +13,10 @@ type Props = {
 };
 
 export const ConfigStep: React.FC<Props> = ({ fields, setFields }) => {
-  const [selectedValue, setSelectedValue] = useState<string | undefined>(fields.model);
+  const [selectedModelValue, setSelectedModelValue] = useState<string | undefined>(fields.model);
+  const [selectedTempValue, setSelectedTempValue] = useState<number | undefined>(
```
EugeneLightsOn, line 17: NIT: Do you mind to rename setSelected**Temp**Value to setSelectedTemperatureValue

EugeneLightsOn (approved):
Great PR, Eric! Just one small nit, and it’s ready to go.

src/interfaces/assistants_web/src/components/AgentSettingsForm/ConfigStep.tsx
ezawadski: Not at all! I have no problem with verbose names!

tianjing-li (changes requested):
Seems the backend already had a temperature value on Agents.

Did you check that the Agent's updated temperature value gets set in the chat_request, see `process_chat()`

I think you might need to set it like these lines

```
# Set the agent settings in the chat request
chat_request.model = agent.model
chat_request.preamble = agent.preamble
```

Otherwise frontend changes lgtm

ezawadski:
> Seems the backend already had a temperature value on Agents.
> 
> Did you check that the Agent's updated temperature value gets set in the chat_request, see `process_chat()`
> 
> I think you might need to set it like these lines
> 
> ```
> # Set the agent settings in the chat request
> chat_request.model = agent.model
> chat_request.preamble = agent.preamble
> ```
> 
> Otherwise frontend changes lgtm

@tianjing-li

The backend uses the temperature passed with the ChatRequest when the chat endpoint is called. I think this is better then defaulting to what is set on the Model because it allows a user to potentially change the temperature between messages. I have the frontend take the value from the model and pass it with the request to the backend.

We keep this business logic in the frontend and leave the backend more flexible and scalable.

tianjing-li:
@ezawadski I believe the opposite might happen if we don't enforce these values being used by the backend. 

We could send `temperature=0.8` in the request body to `POST /chat-stream` for example while the Agent actually has a temperature set to `0.5`, and it would use the `0.8` value rather than the one on the Agent, this is my understanding at least. We can chat more next week online if needed

tianjing-li (approved):
Looks great thank you

PR #918, opened 2025-01-21
backend: (Part 1) Use DB config values for Deployments during runtime

tianjing-li:
I noticed while working on https://linear.app/cohereai/issue/TLK-2601/add-update-to-db-config-when-app-starts that in fact the DB values for deployment configs weren't being used at all during runtime. We were instead always defaulting to what was stored in our Pydantic settings.

This PR modifies core behaviour so we store actual keys that reflect what each deployment will fetch in the newly named
`get_deployment_config_var` method.

**AI Description**

This PR introduces a new `DeploymentSettingsMixin` class in `src/backend/config/settings.py`, which formats deployment configurations before saving them to the database. The mixin class includes a `to_dict` method that retrieves capitalized variable names from a list of strings and returns a new dictionary with these names as keys.

The PR also renames several functions and classes to include the word "instance" in their names, indicating a shift towards working with specific instances of deployments rather than general deployment concepts. This change is reflected in the following files:
- `src/backend/chat/custom/custom.py`
- `src/backend/chat/custom/utils.py`
- `src/backend/config/settings.py`
- `src/backend/crud/deployment.py`
- `src/backend/database_models/deployment.py`
- `src/backend/model_deployments/azure.py`
- `src/backend/model_deployments/bedrock.py`
- `src/backend/model_deployments/cohere_platform.py`
- `src/backend/model_deployments/sagemaker.py`
- `src/backend/model_deployments/single_container.py`
- `src/backend/model_deployments/utils.py`
- `src/backend/routers/utils.py`
- `src/backend/schemas/deployment.py`
- `src/backend/services/deployment.py`
- `src/backend/services/request_validators.py`
- `src/backend/tests/unit/services/test_deployment.py`

Additionally, the PR adds a new function, `get_deployment_by_class_name`, to `src/backend/crud/deployment.py`, which retrieves a deployment by its class name.


EugeneLightsOn (approved):
Looks great.

PR #922, opened 2025-01-24
Backend: Fix DB session not being patched by test fixture

ezawadski:
Noticed that the session in the `get_deployment` method was not being patched by the fixture. Decided to pass request session through function calls to ensure the same session is being used for all db requests.

**AI Description**

This PR introduces a new `session` parameter to the `get_deployment` function in src/backend/chat/custom/utils.py, which is then used in various other functions and methods across the codebase.

- The `get_deployment` function now takes an additional `session` parameter of type `Session`.
- The `session` parameter is then passed to the `deployment_service.get_deployment_instance_by_name` function, replacing the previous `session = next(get_session())` line.
- The `CustomChat.chat` method in src/backend/chat/custom/custom.py now takes a new `session` parameter, which is used in the `get_deployment` call and the `self.call_chat` method.
- The `call_chat` method in the same file also receives a new `session` parameter, replacing the previous `session = kwargs.get("session")` line.
- The `chat` function in src/backend/routers/chat.py and src/backend/services/conversation.py now passes the `session` parameter to the `CustomChat.chat` method.
- The `search_conversations` function in src/backend/routers/conversation.py now uses the `session` parameter in the `get_deployment` call.

### DIFF
--- a/src/backend/chat/custom/custom.py
+++ b/src/backend/chat/custom/custom.py
@@ -1,6 +1,7 @@
 from typing import Any, AsyncGenerator, Dict, List
 
 from fastapi import HTTPException
+from sqlalchemy.orm import Session
 
 from backend.chat.base import BaseChat
 from backend.chat.custom.tool_calls import async_call_tools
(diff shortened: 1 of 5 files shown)


tianjing-li (approved):
Lgtm

EugeneLightsOn (approved):
Lgtm

PR #929, opened 2025-01-28
Tools: sharepoint tool

ezawadski:
Added Sharepoint Tool

**AI Description**

This PR introduces the Sharepoint tool, which allows users to search and retrieve documents from Sharepoint. The tool is integrated into the backend and frontend, providing a seamless experience for users.

## Changes:
- **New Sharepoint Tool**: The PR adds a new tool called Sharepoint, enabling users to search and access documents stored in Sharepoint.
- **Sharepoint Tool Setup**: A new document provides detailed instructions for setting up the Sharepoint tool, including configuring API access, registering a new application, and setting permissions.
- **Sharepoint Configuration**: The `configuration.template.yaml` and `secrets.template.yaml` files are updated to include Sharepoint-specific configurations, such as `tenant_id`, `client_id`, and `client_secret`.
- **Sharepoint Settings**: A new class, `SharepointSettings`, is introduced in `settings.py` to handle Sharepoint-specific settings.
- **Sharepoint Tool Integration**: The Sharepoint tool is integrated into the backend and frontend, with relevant imports and class definitions added to various files, including `__init__.py`, `tool.py`, `auth.py`, and `constants.py`.
- **Sharepoint Authentication**: The `auth.py` file implements the `SharepointAuth` class, which handles authentication for the Sharepoint tool, ensuring secure access to user data.
- **Sharepoint Constants**: The `constants.py` file defines constants specific to the Sharepoint tool, such as `SHAREPOINT_TOOL_ID` and `SEARCH_LIMIT`.
- **Sharepoint Tool Implementation**: The `tool.py` file contains the `SharepointTool` class, which defines the tool's functionality, including searching for documents, retrieving file contents, and serializing results.
- **Sharepoint Utilities**: The `utils.py` file provides utility functions for the Sharepoint tool, such as `serialize_metadata` and `serialize_file_contents`, which handle data serialization and file content processing.
- **Sharepoint Connection Component**: A new component, `SharepointConnection`, is added to the frontend, allowing users to connect to Sharepoint, authenticate, and manage their connection.
- **Sharepoint Icon**: A new icon, `Sharepoint`, is introduced in the frontend, providing a visual representation of the Sharepoint tool.


src/interfaces/assistants_web/src/assets/icons/Sharepoint.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { SVGProps } from 'react';
+
+import { cn } from '@/utils';
+
+export const Sharepoint: React.FC<SVGProps<SVGSVGElement>> = ({ className, ...props }) => (
+  <svg viewBox="52 42 88 66" fill="evenodd" className={cn('h-full w-full', className)} {...props}>
```
ezawadski, line 7: This needs to be replaced. @EugeneLightsOn who would be the best way to generate a SVG with the Sharepoint Logo?

EugeneLightsOn: I'll provide it to you

src/backend/tools/sharepoint/tool.py
```diff
@@ -0,0 +1,158 @@
...
+            result = {}
+            if (resource := hit.get("resource")) is not None:
+                result.update(**serialize_metadata(resource))
+
+            content = serialize_file_contents(content, result.get("name", ""))
+            result.update({
+                "text": content,
+            })
+            results.append(result)
+
```
EugeneLightsOn, line 157: Let's return no results error here 
```
 if not results:
            logger.info(event="[Sharepoint] No documents found.")
            return self.get_no_results_error()
```

src/backend/tools/sharepoint/tool.py
```diff
@@ -0,0 +1,158 @@
...
+        if not response.ok:
+            error = body.get("error", {})
+            error_code = error.get("code")
+            error_description = error.get("message")
+            error_message = f"{error_message}: {error_code}. {error_description}"
+            logger.error(event=f"[Sharepoint] Search error: {error_message}")
+            raise  Exception(error_message)
+
+        return body["value"][0]["hitsContainers"]
+
```
EugeneLightsOn, line 103: Let's check that `body` is not empty

EugeneLightsOn (commented):
Looks good to me - just a few small changes requested

docs/custom_tool_guides/sharepoint.md
```diff
@@ -0,0 +1,39 @@
...
+
+Under the newly registered application navigate to the `Certificates & secrets` page. Click `New client secret`, enter a description and an expiry then click `Add`. Your new Client Secret is only available to copy under the `value` column of the table right now. Copy it into the `secrets.yaml` file in the config directory of the backend.
+
+## 5. Run the Backend and Frontend
+
+run next command to start the backend and frontend:
+
+```bash
+make dev
+```
```
EugeneLightsOn, line 39: Please add the link to the main README file.

src/backend/tools/tavily_search.py
```diff
@@ -37,11 +37,11 @@ def get_tool_definition(cls) -> ToolDefinition:
...
             category=ToolCategory.WebSearch,
             description="Returns a list of relevant document snippets for a textual query retrieved from the internet.",
-        )
+        ) # type: ignore
 
     async def call(
-        self, parameters: dict, ctx: Any, session: DBSessionDep, **kwargs: Any
-    ) -> List[Dict[str, Any]]:
+        self, parameters: dict, ctx: Context, **kwargs: Any,
+    ) -> list[dict[str, Any]]:
```
EugeneLightsOn, line 44: The `session` positional argument is removed here, but in the hybrid search, it is passed positionally, which will cause an error.

EugeneLightsOn (changes requested):
Looks good, small fix is requested

src/backend/tools/tavily_search.py
EugeneLightsOn: We need to check this for all search tools.

ezawadski: There are some tools that use it, some don't. In some cases they pull it from the kwargs, other cases the use it as a positional argument. I removed it here since it was not used, and so that it matches the parent abstract method.

ezawadski: What do you want to do? move it to kwargs in all cases? Or add the session as a positional arg in all cases? Or override the abstract method we the additional positional arg only when it's used?

ezawadski: In Hybrid search I could add it to the kwargs? That's the simplest easy solution.

EugeneLightsOn: I think moving it to kwargs in all cases where it used will be good enough.

EugeneLightsOn (approved):

PR #935, opened 2025-02-05
Backend: include required headers in swagger

ezawadski:
Adjusted the injectable dependencies to specify the headers expected.

Had to change organization_id as a query parameter because it was conflicting with organization-id header in the generated client. This had no effect on frontend, because organization_id query parameter is not used.

Also took care of TODO for having auth on only some endpoints in the user router.

**AI Description**

This PR introduces a new `DependencyType` class to the `src/backend/config/routers.py` file, which is used to define the dependencies for each router. The `RouterName` class is also updated to include a new `DependencyType` attribute.

The `src/backend/main.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/routers/agent.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/routers/organization.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/routers/user.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/services/auth/request_validators.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/services/auth/utils.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/services/request_validators.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/integration/routers/test_agent.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/integration/routers/test_conversation.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/unit/routers/auth/test_authorization_header.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/unit/routers/test_chat.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/unit/routers/test_conversation.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/backend/tests/unit/services/auth/test_request_validators.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/interfaces/assistants_web/src/cohere-client/generated/services.gen.ts` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/interfaces/coral_web/src/cohere-client/generated/services.gen.ts` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/interfaces/coral_web/src/cohere-client/generated/types.gen.ts` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

The `src/interfaces/assistants_web/src/components/UI/Slider.tsx` file is updated to import the new `DependencyType` class and use it to set the router dependencies.

- The `src/backend/config/routers.py` file is updated to include a new `DependencyType` class, which is used to define the dependencies for each router. The `RouterName` class is also updated to include a new `DependencyType` attribute.
- The `src/backend/main.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/routers/agent.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/routers/organization.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/routers/user.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/services/auth/request_validators.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/services/auth/utils.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The `src/backend/services/request_validators.py` file is updated to import the new `DependencyType` class and use it to set the router dependencies.
- The <co: 8>`src/backend/tests/integration


src/backend/services/request_validators.py
```diff
@@ -116,45 +118,46 @@ def validate_user_header(session: DBSessionDep, request: Request):
...
 
 
-def validate_organization_header(session: DBSessionDep, request: Request):
+def validate_organization_header(
+    session: DBSessionDep,
+    Organization_Id: Annotated[str|None, Header( # Not following snake_case so FastAPI parses dependencies properly
+        title="Organization ID",
+        description="Unique Identifier for the Organization making the request",
+    )] = None,
+) -> None:
```
EugeneLightsOn, line 127: 👍

src/backend/services/request_validators.py
```diff
@@ -90,20 +92,20 @@ def validate_deployment_config(deployment_config, deployment_db):
...
 
-    Args:
-        request (Request): The request to validate
-
     Raises:
         HTTPException: If no `User-Id` header.
-
     """
-
-    user_id = get_header_user_id(request)
```
EugeneLightsOn, line 106: Is it okay that we no longer check user_id using the Authorization header?

EugeneLightsOn (approved):
Looks good to me—just one small question to confirm

src/backend/services/request_validators.py
ezawadski: We only use this validator when we are checking for the User-Id Header. And FastAPI pulls it from the header for us by using the Anotated Header type in the function args. That's why we don't need to use the method. This could actually hide bugs when we are supposed to be using the User-Id header instead of the Authorization Header, because the get_header_user_id checks both headers.

EugeneLightsOn: Thanks  for the clarification.

